Issue #25, opened 2016-05-09
Queued Requests are not logged with the x-queue Header


mcweba:
The queued requests are not logged with the x-queue Header. This makes it impossible to identify them in the log files. 
Find a way to log the request before it is queued and the x-queue header gets removed (QueuingHandler, line 57)

**Solution**

1. In the QueuingHandler don't remove the x-queue header, but set an additional header eg. x-queue-inbound-processed="true"
2. In the QueuingHandler the method isQueued will check additionally if the the x-queue-inbound-processed is **not** set

lbovet:
This is quite a change in the semantics introducing incompatibility. This will break some of our use cases because we forward queued requests to other queues. We will have to remove this new header.
I would prefer to let this be done in client code (this is a logging specific feature), clients code can just make a copy of the x-queue header into a custom header for logging purpose.
Headers are part of the gateleen API. If their semantic changes, the API changes.

mcweba:
As discussed with @lbovet, we don't add this functionality to gateleen itself. To log this header anyway, an additional header has to be added to the request which then will be logged. See example code below:

```java
private void addQueueLogHeader(HttpServerRequest request){
	String queue = request.getHeader("x-queue");
	if(queue != null){
		request.headers().set("x-queue-log", queue);
	}
}
```
This issue will be closed without modification.

PR #26, opened 2016-05-09
#25 Queued Requests are not logged with the x-queue Header

mcweba:
Please review and eventually merge

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ This is a list of the custom headers used by Gateleen.
 | x-client-timestamp | timestamp of the client while sending the request | |
 | x-server-timestamp | timestamp of the server while sending the request | |
 | x-queue            | indicates a queued request | [gateleen-queue](gateleen-queue/README_queue.md) |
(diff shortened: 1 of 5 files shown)


lbovet:
This is quite a change in the semantics introducing incompatibility. This will break some of our use cases because we forward queued requests to other queues. We will have to remove this new header.
I would prefer to let this be done in client code (this is a logging specific feature), clients code can just make a copy of the x-queue header into a custom header for logging purpose.
Headers are part of the gateleen API. If their semantic changes, the API changes.

mcweba:
This pull request will be closed without merging to not break existing functionality. To allow the logging of the x-queue header, an additional header can be added in the verticle of the instance using gateleen.

The code to add such a header could look something like this:
```java
private void addQueueLogHeader(HttpServerRequest request){
	String queue = request.getHeader("x-queue");
	if(queue != null){
		request.headers().set("x-queue-log", queue);
	}
}
```

Issue #27, opened 2016-05-12
Introduce randomness in scheduler to avoid network load peaks


ljucam:
Schedulers user cron expressions and are thus aligned with the clock. That means that all Gateleen servers with the same configuration will performs scheduled requests at the same time.
For requests going to a central communication server, this will introduce unbearable network load peaks.
We should add a randomness configuration value. This defines the maximal offset in seconds.

For example: 
```
    randomOffset: 300 
```
means that the actual time can be shifted up to 5 minutes after the planned time.

**Summary:** 
•	We add a new configuration property to define the maximal offset in seconds.
•	During the load of the scheduler configuration we calculate an offset for each scheduler, which stays the same for the specific schedulers, till the next reload of the scheduler configuration. This way the interval between executions of the schedulers remain always the same.

PR #28, opened 2016-05-13
fixed Introduce randomness in scheduler to avoid network load peaks #27

ljucam:


gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
```diff
@@ -36,24 +35,48 @@
...
+
+    /**
+     * Calculates a randomOffset (delay) before a scheduler
+     * is triggered. The possible offset range is limited bx the
+     * maxRandomOffeset. If the maxRandomOffset is 0, no calculation
+     * is performed.
+     *
+     * @param maxRandomOffset possible offset range in seconds
+     */
+    private void caltRandomOffset(int maxRandomOffset) {
```
lbovet: Should be "calc"

PR #37, opened 2016-06-02
Connection handling should not be forwarded

lbovet:


floriankammermann:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.10 we should also remove any headers fields with the same name as the connection-token (remove any header field(s) from the message with the same name as the connection-token.)

PR #39, opened 2016-06-03
fixed #38 Scheduler doesn't set "X-Server-Timestamp" header

### LINKED ISSUE #38: Scheduler doesn't set "X-Server-Timestamp" header
The Scheduler doesn’t currently set a ‘X-Server-Timestamp’ header when it enqueues a request. 
Because of this the expiration of schedulers queues is broken. 

Add the "X-Server-Timestamp" header to each request of a schedulers, when it is enqueued.

ljucam:
fixed #38 Scheduler doesn't set "X-Server-Timestamp" header


ljucam:
mixed with wrong branch

PR #62, opened 2016-07-12
Implement a ZipExtractHandler in the Expansion module

ljucam:


lbovet:
Is the mime type (content-type) of the extracted resource set? I guess it should be a must.

PR #67, opened 2016-07-19
DELETE requests should not block block queue if "404 NOT FOUND"

oli-h:
That is: the target resource:
- is already deleted (for whatever reason)
- never existed

I.o.w: If target server answers "404 NOT FOUND" for a DELETE request,
then we accept this and dequeue the DELETE request.

### DIFF
--- a/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
+++ b/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
@@ -6,6 +6,7 @@ import io.vertx.core.buffer.Buffer;
 import io.vertx.core.eventbus.Message;
 import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpClientRequest;
+import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


ljucam:
There already exists a class which is responsible for the translation.
You can found it here: _org.swisspush.gateleen.routing.Translator_

By adding the header ‘"x-translate-status-4xx" : "200"’ you get a translation of every 4xx status code to the status code 200.

lbovet:
Closing, since already possible from client and hooks should set this header.
If there are cases where it does not work as expected, please open an issue.

oli-h:
I do not agree:

1. The x-translate-status does not distinguish between http methods. Here i just want to ignore 404 on a DELETE request
2. Queues can be used without hooks. Services can directly PUT/DELETE with "x-queue" header set. If any one misses to set the "x-translate-header" on a DELETE this will freeze the queue.
3. In the modified class QueueProcessor.java there is already a special handling for "409 CONFLICT" http responses. So one might ask why - as this could also be solved with x-translate-status.
4. A log entry, level "WARN" is written when this "i accept 404 on DELETE" occurs. The x-translate-status does not log anything. But a DELETE on a non existing resource indicates some process error elsewhere - so we want to know about
5. Gateleen-Queue could probably be used without Gateleen-Router - but the x-translate-status is only activated when the request passes the router

lbovet:
These comments are valid if we consider that Gateleen is responsible to avoid blocked queues. In fact, blocking queues is currently a feature of gateleen. This forces wrong-behaving clients to fix their bugs. The current startegy is to delegate the decision to block (or not) the queue to the client. Gateleen is meant to be sufficiently generic to let users implement their own behaviour (thus, I agree that the hardcoded 409 handling is not in-line with this).
As long as we consider the client as responsible of the queue, I would not change this behaviour. Hooks are responsible of their queues as well and that's why they choose to translate status.
Point 4.: a 404 on DELETE can be some process error but it can also be an expected result in normal operation. Gateleen should not make assumption on these semantics, again the client is responsible.
Point 5. Actually the processed request must be forwarded somewhere, so it is always routed. We preferred to have one unique place where we translate the status and this is conceptually correct to do it in the forwarder. 
Queues gives a promise to the client. Either the request worked, or it will be kept in the queue. This is the default. If you are ok with discarding 4xx requests, you have to tell it explicitely.
Note that with routing rules including method matching and static header, you can provide that transparently to the clients, if we want to avoid the situation where clients forget to pass the translate header.

PR #68, opened 2016-07-19
Implement a DelegateHandler as a new module fixes #59 

### LINKED ISSUE #59: Implement a DelegateHandler as a new module
Implement the delegate feature. With this mechanism we are able to replace a request by another. 
For example, we could make a GET that will be replaced by another one (or even a POST). This allows for creating cheap adapters.

Cookbook for the delegate: 
Name a delegate: 
`/gateleen/server/delegate/v1/delegates/<delegate>/`

Create a definition:
`/gateleen/server/delegate/v1/delegates/<delegate>/definition`

DelegateHandler automatically checks if the Pattern 
`/gateleen/server/delegate/v1/delegates/<delegate>/execution`
is called. 

Example: 
```
 // create a delegate
PUT /gateleen/server/delegate/v1/delegates/user-zip-copy/definition
{
    "methods": [ "PUT", "DELETE" ],
    "pattern": ".*/([^/]+.*)",
    "requests": [
        {
            "method": "POST",
            "url": "/gateleen/server/v1/copy",
            "payload": {
                "source": "/gateleen/$1?expand=100&zip=true",
                "destination": "/gateleen/zips/users/$1.zip"
            }
        }
        ]
    }
}

```

```
// create a hook triggering the delegate
PUT /gateleen/users/_hooks/listeners/user-zip-copy
{
    "destination": "/gateleen/server/delegate/v1/delegates/user-zip-copy/execution"
}
```

ljucam:


gateleen-delegate/src/main/java/org/swisspush/gateleen/delegate/Delegate.java
```diff
@@ -0,0 +1,208 @@
...
+            private AtomicInteger currentIndex = new AtomicInteger(0);
+
+            @Override
+            public void handle(HttpClientResponse response) {
+                if ( LOG.isTraceEnabled() ) {
+                   LOG.trace("Done handler - handle");
+                }
+
+                // request was fine
+                if ( response.statusCode() == StatusCode.OK.getStatusCode() ||
```
lbovet: Here we should accept all 2xx response codes, not only 200 and 202. We may have a 204, for example.

gateleen-delegate/src/main/java/org/swisspush/gateleen/delegate/DelegateHandler.java
```diff
@@ -0,0 +1,366 @@
...
+
+    /**
+     *
+     * Tries to extract the name of the delegate out
+     * of the incoming uri.
+     *
+     * @param uri original request uri
+     * @return the name of the delegate or null if nothing matches
+     */
+    private String getDelegegateName(final String uri) {
```
lbovet: Delegegegegegeate

lbovet:
Good job, mate!

Issue #73, opened 2016-07-25
Add property to execute scheduler on startup


thwint:
Currently the schedulers are not executed at startup. If a scheduler only runs after a long period and the system is restarted before this period is over it never gets executed.

To ensure that a job is running at least once it should be possible to create a scheduler which is executed at startup.

Therefore I propose to introduce a new property.

Proposal:
"executeOnStartup" : true

Please comment.

PR #75, opened 2016-07-25
#73 Add property to executescheduler on startup

thwint:
Added property executeOnStartup

### DIFF
--- a/gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
+++ b/gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
@@ -95,16 +95,11 @@ public class Scheduler {
                     }
                 });
             }
-            if(!executed && executeOnStartup ){
-                executed = true;
-                vertx.setTimer((randomOffset + 1) * 1000, new Handler<Long>() {
-                    @Override
-                    public void handle(Long aLong) {
-                        trigger();
-                    }
-                });
-            }
         }));
+        if(!executed && executeOnStartup ){
+            executed = true;
+            trigger();
+        }
     }
 
     public void stop() {
--- a/gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
+++ b/gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
@@ -37,10 +37,12 @@ public class Scheduler {
     private long timer;
     private MonitoringHandler monitoringHandler;
     private long randomOffset = 0L;
+    private boolean executeOnStartup=false;
+    private boolean executed = false;
 
     private Logger log;
 
-    public Scheduler(Vertx vertx, RedisClient redisClient, String name, String cronExpression, List<HttpRequest> requests, MonitoringHandler monitoringHandler, int maxRandomOffset) throws ParseException {
+    public Scheduler(Vertx vertx, RedisClient redisClient, String name, String cronExpression, List<HttpRequest> requests, MonitoringHandler monitoringHandler, int maxRandomOffset, boolean executeOnStartup) throws ParseException {
         this.vertx = vertx;
         this.redisClient = redisClient;
         this.name = name;
@@ -49,6 +51,7 @@ public class Scheduler {
         this.log = LoggerFactory.getLogger(Scheduler.class.getName()+".scheduler-"+name);
         this.monitoringHandler = monitoringHandler;
         calcRandomOffset(maxRandomOffset);
+        this.executeOnStartup=executeOnStartup;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
```diff
@@ -92,6 +95,15 @@ public void start() {
                     }
                 });
             }
+            if(!executed && executeOnStartup ){
+                executed = true;
+                vertx.setTimer((randomOffset + 1) * 1000, new Handler<Long>() {
```
lbovet: We should not have randomOffset here. trigger should be called directly

gateleen-test/src/test/java/org/swisspush/gateleen/scheduler/SchedulerTest.java
```diff
@@ -169,6 +169,108 @@ public void testSchedulerWithRandomOffset(TestContext context) throws Exception
...
+        String schedulerName = "scheduler-executeOnStartup-test";
+        String resourceName = "schedulerWithExecuteInStartup";
+        String cronExpression = "* * 1/1 * * ?";
+
+        // scheduler config
+        String testSchedulers = "{\n"
+                + "  \"schedulers\": {\n"
+                + "    \"" + schedulerName + "\": {\n"
+                + "      \"cronExpression\": \"" + cronExpression + "\",\n"
+                + "      \"executeOnStartup\": " + true + ",\n"
```
lbovet: You can write true in the string directly

gateleen-test/src/test/java/org/swisspush/gateleen/scheduler/SchedulerTest.java
```diff
@@ -169,6 +169,108 @@ public void testSchedulerWithRandomOffset(TestContext context) throws Exception
...
+        // wait for the scheduler to be created
+        await().atMost(10, SECONDS).until( () -> {
+            if ( schedulerResourceManager.getSchedulers() != null ) {
+                return schedulerResourceManager.getSchedulers().stream().filter(s -> s.getName().equals(schedulerName)).findFirst().get();
+            }
+            return null;
+        }, is(notNullValue()));
+
+        // get the scheduler and the randomOffset
+        Scheduler scheduler = schedulerResourceManager.getSchedulers().stream().filter(s -> s.getName().equals(schedulerName)).findFirst().get();
```
lbovet: Where is then "scheduler" used?

gateleen-test/src/test/java/org/swisspush/gateleen/scheduler/SchedulerTest.java
```diff
@@ -169,6 +169,108 @@ public void testSchedulerWithRandomOffset(TestContext context) throws Exception
...
+    @Test
+    public void testSchedulerWithExecuteOnStartupTrue(TestContext context) throws Exception {
+        Async async = context.async();
+        delete();
+
+        // prepare the settings
+        String content = String.valueOf(System.currentTimeMillis());
+        String schedulerName = "scheduler-executeOnStartup-test";
+        String resourceName = "schedulerWithExecuteInStartup";
+        String cronExpression = "* * 1/1 * * ?";
```
lbovet: How often does this cron expression runs?

gateleen-test/src/test/java/org/swisspush/gateleen/scheduler/SchedulerTest.java
```diff
@@ -169,6 +169,108 @@ public void testSchedulerWithRandomOffset(TestContext context) throws Exception
...
+                return schedulerResourceManager.getSchedulers().stream().filter(s -> s.getName().equals(schedulerName)).findFirst().get();
+            }
+            return null;
+        }, is(notNullValue()));
+
+        // get the scheduler and the randomOffset
+        Scheduler scheduler = schedulerResourceManager.getSchedulers().stream().filter(s -> s.getName().equals(schedulerName)).findFirst().get();
+
+        // wait for the vertx timer to fire the first time
+        await().atMost(10, SECONDS).until( () -> jedis.get("schedulers:" + schedulerName), is(notNullValue()));
```
lbovet: Here you test that the scheduler has run...

gateleen-test/src/test/java/org/swisspush/gateleen/scheduler/SchedulerTest.java
```diff
@@ -169,6 +169,108 @@ public void testSchedulerWithRandomOffset(TestContext context) throws Exception
...
+
+        // Test if first run
+        await().atMost(30, SECONDS).until(() -> get("/tests/" + SERVER_NAME + "/" +resourceName ).then().extract().statusCode(), equalTo(404));
```
lbovet: ...and here you test that it had no effect. I don't see that as OK.

gateleen-scheduler/src/main/java/org/swisspush/gateleen/scheduler/Scheduler.java
```diff
@@ -92,6 +95,15 @@ public void start() {
                     }
                 });
             }
+            if(!executed && executeOnStartup ){
```
lbovet: Why is this in the setPeriodic? We should just call trigger() inside start, no?

thwint:
Thank you for this review. You are absolutely right. I will correct this. 
Only one question concerning the tests. 
We don't want to test if the schedulers are executed periodically. We already do this within other tests. But we want to ensure that in one case the scheduler is triggered at startup and in the other case it is not.
Do you have a proposal how to test that it is ok?

PR #120, opened 2017-02-23
Use right default Gateleen port 7012 for Playground upload.

Kusig:
The default port of Gateleen is 7012 and therefore the playground
resource upload should be done to this port as well by default.


ljucam:
we try again

Kusig:
Don't know why this happened. It somehow merged with another branch which exists directly on the master.

PR #160, opened 2017-08-14
provides the hook feature to angular js applications

dominik-cnx:


dominik-cnx:
also plan to provide selenium tests against the UI in the future

PR #179, opened 2017-11-20
Gapa integration example

brunm1:
Add line in request handler to record inbound requests with Gapa
Add line in Forwarder handler to record outbound requests with Gapa
The gapa integration module is not yet added to the dependencies.

Note that this pull request is only an example to show where some lines are added. I didn't change the gradle build configuration.


lbovet:
I close the PR but we can still refer to it for future Gapa-related work

PR #191, opened 2018-01-10
Update to Vert.x 3.5.0

oli-h:
This includes update for RedisQues 2.5.0 and rest-storage 2.5.1 which also are based on Vert.x 3.5.0

Additonal work:
- cleanup POMs and dependencies
- introduced module "gateleen-testhelper" which contains base classes and Maven dependencies for Unittests. Previously they were part of gateleen-core _productive_ code - this leaded to a "compile" (and not "test") dependency to JUnit, Jedis, Mockito etc...


oli-h:
Tests on Travis-CI failed probably due to slow execution.
Failed test step: https://github.com/swisspush/gateleen/blob/develop/gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteTest.java#L210

Here, the test expects within 2 seconds a specfic outcome. Tests on localhost are stable.
Any idea? Increasing here the timeout does probably not solve the problem sustainly as a lot of test steps seem to work with such a "within 2 seconds else fail"

PR #194, opened 2018-01-26
some logs removed, loglevel changed to debug

dominik-cnx:


dominik-cnx:
Reduces noise in the console

PR #198, opened 2018-02-15
Issue_192

bertranoel:
Bugfix for wrong usage of request.header() instead of request.response().headers().
Also centralized this block of operations in an util class.


oli-h:
Note: headers().setAll(xxx) clears the headers before setting them from xxx.
In other words: headers added _before_ calling setAll(...) are lost. Are you aware of this?

PR #201, opened 2018-02-20
Gateleen issue 43: provide dynamic Header manipulations for Reouting and Hooking

### LINKED ISSUE #43: Extend staticHeaders functionality
With the current implemenation we can define static headers for routing rules like this:

``` json
{
    "/playground/tests/(.*)": {
        "path": "/playground/server/tests/$1",
        "storage": "main",
        "staticHeaders": {
            "X-Expire-After": "3600"
        }
    }
}
```

Defining a _X-Expire-After_ static header like in the example above will add this header to the request. However, an already added _X-Expire-After_ header will be **replaced**.

It would be useful to have a feature which allows to add a static headers **only when they are not already present**.

This could be accomplished by adding a **forced** flag. This flag would result in the following behaviour:

| Force flag | Behaviour |
| --- | :-- |
| true | Replaces (overrides) already defined header values with the staticHeader value |
| false | Does nothing when header value already exists, adds defined header value otherwise |
| not set | Replaces (overrides) already defined header values with the staticHeader value. This is used for backward compatibility, since headers are replaced in current implementation |

A possible staticHeader configuration could look like this:

``` json
{
    "/playground/tests/(.*)": {
        "path": "/playground/server/tests/$1",
        "storage": "main",
        "staticHeaders": {
            "X-Expire-After": { "value": "3600", "force": true }
        }
    }
}
```

oli-h:
(closes #43)

- core functionality in gateleen-core (HeaderFunctions....), including Unittest
- used in gateleen-routing and gateleen-hook
- adapt routing rules JSON schema (additional property "headers" - property "staticHeaders" is now deprecated)
- Hooks are now also schema validated- that's why we have lots of changes in HookHandler.java

Note to JSON Schema validation (while placing Hooks): Uses a different JSON Schema/Validation library instead of gateleen-validator. Reasons:
1. gateleen-validator is based on FGE's json-schema-validator (https://github.com/java-json-tools/json-schema-validator/)
   This project is stale since years (last remarkable activity January 2014)
2. NetworkNT's validator is 10 times faster than FGE's one (see JsonSchemaValidationPerformanceCompare.java)
3. gateleen-validator does no offer API to pre-parse and reuse schema instance. Schema has to be parsed for every validation run
   This is another performance issue
As we now wat to validate every PUT of a hook we must be more aware of performance - so we take another approach and I don't want to refactor gateleen-validator at the moment.


oli-h:
Bump....

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HeaderFunctions.java
```diff
@@ -0,0 +1,239 @@
...
+     *
+     * @param config the JSON configuration for the header manipulator chain
+     */
+    public static HeaderFunction parseFromJson(JsonArray config) throws IllegalArgumentException {
+        LOG.debug("creating header function chain from " + config);
+
+        Consumer<EvalScope> chain = null;
+        for (int pos = 0; pos < config.size(); pos++) {
+            JsonObject rule = config.getJsonObject(pos);
+            Consumer<EvalScope> c = parseOneFromJason(config, rule);
```
lbovet: Json, not Jason :)

oli-h: next time: Tschaisonn

lbovet: Djézonne, in french

PR #204, opened 2018-02-27
[ISAGD-4215] Dont Forward 'Connection' headers.

hiddenalpha:
According to RFC 2616 section 14.10
(https://tools.ietf.org/html/rfc2616#section-14.10)

The Connection headers "MUST NOT be communicated by proxies over further
connections".

This was already implemented earlier in the request stream in gateleen.
But not for the response stream. This will implement it for this stream
now.

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -18,8 +18,8 @@ public class HttpHeaderUtil {
      * @param headers
      *      The headers to check.
      */
-    public <T extends MultiMap> T removeNonForwardHeaders( T headers ) {
-        final String CONNECTION = HttpRequestHeader.CONECTION.getName();
+    public static <T extends MultiMap> T removeNonForwardHeaders( T headers ) {
+        final String CONNECTION = HttpRequestHeader.CONNECTION.getName();
 
         // Remove all headers named by connection-token.
         headers.getAll(CONNECTION).forEach(headers::remove);
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -1,0 +1,33 @@
+package org.swisspush.gateleen.core.util;
+
+import io.vertx.core.MultiMap;
+
+
+public class HttpHeaderUtil {
+
+    /**
+     * <p>Removes headers which MUST NOT be forwarded by proxies.</p>
+     *
+     * <ul>
+     *     <li>This MAY modifies the passed headers.</li>
+     *     <li>This MAY returns the same (modified) instance as passed.</li>
+     * </ul>
+     *
+     * @see <a href="https://tools.ietf.org/html/rfc2616#section-14.10">RFC 2616 section 14.10</a>
+     *
+     * @param headers
+     *      The headers to check.
+     */
+    public <T extends MultiMap> T removeNonForwardHeaders( T headers ) {
+        final String CONNECTION = HttpRequestHeader.CONECTION.getName();
+
+        // Remove all headers named by connection-token.
+        headers.getAll(CONNECTION).forEach(headers::remove);
+
+        // Remove the connection headers itself.
+        headers.remove(CONNECTION);
+
+        return headers;
+    }
+
+}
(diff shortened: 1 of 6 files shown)


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -296,7 +297,14 @@ private HttpClientRequest prepareRequest(final HttpServerRequest req, final Stri
 
             req.response().setStatusCode(statusCode);
 
-            req.response().headers().addAll(cRes.headers());
+            // Add received headers to original request except the 'Connection'
+            // headers. See RFC "https://tools.ietf.org/html/rfc2616#section-14.10".
+            // HINT: This manipulates the originally received headers.
+            final MultiMap headersFromServer = cRes.headers();
+            headersFromServer.getAll("connection").forEach(headersFromServer::remove);
```
mcweba: I think it's implemented a little bit complicated. I made a short unit test:

```java
@Test
public void testMultimap() {
	CaseInsensitiveHeaders params = new CaseInsensitiveHeaders();

	params.add("connection", "abc");
	params.add("test", "123");
	params.add("connection", "def");

	System.out.println("names: " + params.names());
	System.out.println("getall connection: " + params.getAll("connection"));

	params.remove("connection");

	System.out.println("getall connection after remove: " + params.getAll("connection"));
	System.out.println("names after remove: " + params.names());
}
```
Output:
```
names: [connection, test]
getall connection: [abc, def]
getall connection after remove: []
names after remove: [test]
```
So it looks like it could be simply implemented with
```java
req.response().headers().addAll(cRes.headers().remove("connection"));
```
Has to be tested :-)

hiddenalpha: This would not fulfill the specification (see: https://tools.ietf.org/html/rfc2616#section-14.10)

> HTTP/1.1 proxies MUST parse the Connection header field before a
   message is forwarded and, for each connection-token in this field,
   remove any header field(s) from the message with the same name as the
   connection-token. Connection options are signaled by the presence of
   a connection-token in the Connection header field

For example we receiving following headers:

> Connection: close, foo
> Connection: bar, dada
> foo: blubb
> bar: asdf

In this situation we've to remove *all* of these 4 headers.

mcweba: You're right. For the request headers it already was implmented like this. Maybe make a private method to remove these headers, to reduce code duplication

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpRequestHeader.java
```diff
@@ -8,6 +8,7 @@
  * @author https://github.com/mcweba [Marc-Andre Weber]
  */
 public enum HttpRequestHeader {
+    CONECTION("connection"),
```
mcweba: typo => CONNECTION

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -0,0 +1,33 @@
...
+     *     <li>This MAY modifies the passed headers.</li>
+     *     <li>This MAY returns the same (modified) instance as passed.</li>
+     * </ul>
+     *
+     * @see <a href="https://tools.ietf.org/html/rfc2616#section-14.10">RFC 2616 section 14.10</a>
+     *
+     * @param headers
+     *      The headers to check.
+     */
+    public <T extends MultiMap> T removeNonForwardHeaders( T headers ) {
```
mcweba: please make this method static so it can be used like the other util classes

gateleen-core/src/test/java/org/swisspush/gateleen/core/util/HttpHeaderUtilTest.java
```diff
@@ -0,0 +1,47 @@
...
+@RunWith(VertxUnitRunner.class)
+public class HttpHeaderUtilTest {
+
+    private static String CONNECTION = HttpRequestHeader.CONECTION.getName();
+
+    @Test
+    public void removeNonForwardHeadersTest(TestContext testContext) {
+
+        // Prepare instance to test
+        final HttpHeaderUtil httpHeaderUtil = new HttpHeaderUtil();
```
mcweba: not needed when method is static

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
```diff
@@ -84,7 +86,7 @@ public Route(Vertx vertx, ResourceStorage storage, LoggingResourceManager loggin
      * Creates the forwarder for this hook.
      */
     private void createForwarder() {
-        forwarder = new Forwarder(vertx, client, rule, storage, loggingResourceManager, monitoringHandler, userProfilePath);
+        forwarder = new Forwarder(vertx, client, rule, storage, loggingResourceManager, monitoringHandler, userProfilePath, httpHeaderUtil);
```
mcweba: not needed when method is static

PR #205, opened 2018-03-01
[ISAGD-4215] DontForwardConnectionClose

hiddenalpha:
Oops. Shortly reviewed the changes again and saw that I call the utility
method twice.


mcweba:
That would not have happened if you had done it without the variables ;-)

```java
req.response().headers().addAll(HttpHeaderUtil.removeNonForwardHeaders(cRes.headers()));
```

hiddenalpha:
Yes of course. But I prefer code that one can read and is better to understand (one action per line). I think the compiler has no problems optimizing a redundant variable. This is also easier to debug because one can simply comment out a line.

Theoretically we could reduce the whole file to one line. But I think that wouldn't be very helpful ;)

PR #207, opened 2018-03-09
Bugfix/206 houston returns internal server error

hiddenalpha:


gateleen-logging/src/test/java/org/swisspush/gateleen/logging/LoggingResourceManagerTest.java
```diff
@@ -143,6 +148,66 @@ public HttpServerResponse response() {
...
+
+        // Trigger victim to do its work.
+        final boolean returnValue = loggingResourceManager.handleLoggingResource( request );
+
+        // Assert victim accepted our request.
+        testContext.assertTrue( returnValue );
+        // did set a response code,
+        testContext.assertNotNull( responseStatusCode[0] );
+        // responded with a status code of 4xx,
+        testContext.assertTrue( responseStatusCode[0] >= 400 );
```
mcweba: Why not assert that status code is exactly 400? When providing a loggingresource with invalid content I expect the response to be **400 Bad Request**

Issue #208, opened 2018-03-09
Infinite hook expiration


hiddenalpha:
(Moved from ISAGD-4839)

For permanent hooks, we currently need to set a very high value because the usual -1 value does not override the default expiration.

Just test the x-expire-after header before parsing it. If equals to -1, then do not set the hook expiration.

We are talking here about the expiration of the hook itself, not the propagated messages.

hiddenalpha:
I'm working on it.

mcweba:
merged with pull request #209

PR #209, opened 2018-03-12
[#208] infinite hook expiration

hiddenalpha:
Simple implementation for requested feature.


lbovet:
This PR sets the expiration time to a very high values around yr 3000. This is not what is expected.
We need to _not_ set the expiration time. Please change accordingly.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1030,16 +1032,24 @@ private void handleListenerRegistration(final HttpServerRequest request) {
             final String listenerStorageUri = hookRootUri + HOOK_LISTENER_STORAGE_PATH + getUniqueListenerId(request.uri());
 
             // Extract expireAfter from the registration header.
-            Integer expireAfter = ExpiryCheckHandler.getExpireAfter(request.headers());
-            if (expireAfter == null) {
-                expireAfter = DEFAULT_HOOK_STORAGE_EXPIRE_AFTER_TIME;
-            }
+            final String myExpireAfter = request.headers().get( X_EXPIRE_AFTER );
+            final String expirationTime;
+            if( "-1".equals(myExpireAfter) ){
```
mcweba: Why not add this to the ExpiryCheckHandler where we already provide methods for expiry headers? You could provide a method which returns an enum with values like **NEVER**, **EXPIRES_IN_MS** holding the corresponding Integer values. The ExpiryCheckHandler already handles the "-1" case by returning null. You could not return null in this case but the Expiration.NEVER enum value.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1290,6 +1298,31 @@ private void registerListener(Buffer buffer) {
...
+     *
+     * @param expirationTimeExpression
+     *      ISO date to parse.
+     * @return
+     *      The parsed date or null if null got passed.
+     * @throws RuntimeException
+     *      In case of problems while parsing the value. This will most
+     *      probably be an {@link IllegalArgumentException}.
+     */
+    private LocalDateTime parseExpirationTime(String expirationTimeExpression ) {
```
mcweba: Should be moved to ExpiryCheckHandler

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1290,6 +1298,31 @@ private void registerListener(Buffer buffer) {
...
+        LocalDateTime expirationTime;
+        if( EXPIRATION_TIME_NEVER.equals(expirationTimeExpression) ){
+            // No. Not every hook has an expiration time. (See: ISAGD-4839)
```
mcweba: Do not mention Post specific issues like ISAGD-4839. Since this is open source, nobody outside the Post needs this information

hiddenalpha:
Ready to review from my point of view :)

gateleen-hook/pom.xml
```diff
@@ -25,6 +25,12 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
mcweba: Do you need logging in the unit tests?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1032,23 +1041,33 @@ private void handleListenerRegistration(final HttpServerRequest request) {
...
-            if( "-1".equals(myExpireAfter) ){
-                expirationTime = EXPIRATION_TIME_NEVER;
-            }else{
-                Integer expireAfter = ExpiryCheckHandler.getExpireAfter(request.headers());
-                if (expireAfter == null) {
+
+            { // Evaluate expiration time
+                Integer expireAfter;
+                try{
+                    expireAfter = ExpiryCheckHandler.getExpireAfter2( request.headers() );
```
mcweba: The _getExpireAfter2()_ method should not throw exceptions (like the other methods in this class do neither). You could for example return a _java.util.Optional_ object which would be _Optional.absent()_ in error case

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HttpHook.java
```diff
@@ -98,9 +98,9 @@ public void setExpireAfter(int expireAfter) {
     }
 
     /**
-     * Returns the expiration time of this hook.
-     * 
-     * @return expirationTime
+     * @return
```
mcweba: Since this method now can return _null_ values, I would make it more clear to the user, that the value could be _null_. This could be achieved by returning a _java.util.Optinal<LocalDateTime>_ object.

Example:
```java
public Optional<LocalDateTime> getExpirationTime() {
	return Optional.ofNullable(expirationTime);
}
```

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/mocks/FastFailHttpServerRequest.java
```diff
@@ -0,0 +1,204 @@
+package org.swisspush.gateleen.hook.mocks;
```
mcweba: This class is a duplication of the class _org.swisspush.gateleen.core.http.DummyHttpServerRequest_ and can be removed

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/mocks/FastFailHttpServerResponse.java
```diff
@@ -0,0 +1,255 @@
+package org.swisspush.gateleen.hook.mocks;
```
mcweba: This class is a duplication of the class _org.swisspush.gateleen.core.http.DummyHttpServerResponse_ and can be removed

gateleen-hook/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,13 @@
+
```
mcweba: Do you need logging in the unit tests?

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
```diff
@@ -96,7 +96,26 @@ public static void updateServerTimestampHeader(MultiMap headers) {
...
+    }
+
+    /**
+     * Delegates to {@link #getExpireValue(String)}
+     *
+     * This is extended version of {@link #getExpireAfter(MultiMap)} but
+     * handling more cases than the original. Like -1 for infinite or an
+     * Exception for corrupt headers.
+     */
+    public static Integer getExpireAfter2( MultiMap headers ) {
```
mcweba: Please choose a more reasonable name for this method

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
```diff
@@ -108,45 +127,43 @@ public static Integer getExpireAfter(MultiMap headers) {
      * @return queue-expire-after time in seconds or null if nothing is found
      */
     public static Integer getQueueExpireAfter(MultiMap headers) {
-        return getExpireValue(headers, QUEUE_EXPIRE_AFTER_HEADER);
+        try{
```
mcweba: Besides the header property name, it's the exact same implementation as in _getExpireAfter_ which leads to unwanted code duplication

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
```diff
@@ -108,45 +127,43 @@ public static Integer getExpireAfter(MultiMap headers) {
...
-                log.warn(field + " is not a number: " + expireHeaderValue);
-                value = null;
+    private static Integer getExpireValue( final String expirationTimeout ) {
+        final Integer ans;
+        if( expirationTimeout == null ) {
+            // If we get null we'll answer with null.
+            ans = null;
+        }else{
+            // Value exists. Try parsing it (this will throw specified exception on failure).
+            ans = Integer.parseInt( expirationTimeout );
```
mcweba: Maybe rethink exception handling here. In the current implementation you make the exception handling not until _HookHandler_ class. No other util method in this class throws exceptions. We should stick to a single strategy.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1032,23 +1041,33 @@ private void handleListenerRegistration(final HttpServerRequest request) {
...
+                    // Use default if header is not set.
+                    if( expireAfter == null ) expireAfter = DEFAULT_HOOK_STORAGE_EXPIRE_AFTER_TIME;
```
mcweba: To keep a continuous code style across the whole framework, avoid using _single line if statements_

gateleen-queue/src/test/java/org/swisspush/gateleen/queue/mocks/FastFailMultiMap.java
```diff
@@ -0,0 +1,151 @@
+package org.swisspush.gateleen.queue.mocks;
```
mcweba: Move to the _gateleen-core_ package like the _DummyHttpServerRequest_ and _DummyHttpServerResponse_ classes because this class could be useful elsewhere too.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -50,8 +60,23 @@
...
     private HookHandler hookHandler;
 
+    private static final Method registerListenerMethod;
+    private ListenerRepository listenerRepository;
+
+    static {
+        // Lever out access restrictions to make this a true unit test and not an integration
+        // test testing through hundreds of layers.
+        try{
+            registerListenerMethod = HookHandler.class.getDeclaredMethod( "registerListener" , Buffer.class );
```
mcweba: Don't test private methods like this. Just google the reasons why not :-)

A nice quote from the book "Pragmatic Unit Testing":

> In general, you don't want to break any encapsulation for the sake of testing (or as Mom used to say, "don't expose your privates!"). Most of the time, you should be able to test a class by exercising its public methods. If there is significant functionality that is hidden behind private or protected access, that might be a warning sign that there's another class in there struggling to get out.

meaning when you struggle to test the private methods through the public methods, it indicates a possible code smell which should be refactored.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1032,23 +1041,33 @@ private void handleListenerRegistration(final HttpServerRequest request) {
...
             // Extract expireAfter from the registration header.
-            final String myExpireAfter = request.headers().get( X_EXPIRE_AFTER );
             final String expirationTime;
-            if( "-1".equals(myExpireAfter) ){
-                expirationTime = EXPIRATION_TIME_NEVER;
-            }else{
-                Integer expireAfter = ExpiryCheckHandler.getExpireAfter(request.headers());
-                if (expireAfter == null) {
+
+            { // Evaluate expiration time
```
mcweba: I would suggest to extract this block of code in a separate method rather than using an anonymous code block to restrict variable scope.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
hiddenalpha: How then a caller (eg HookHandler:1049) should distinguish between _no-header-using-default_ (Line 1051) and _infinite_ (Line 1060) if both get returned by null?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HttpHook.java
oli-h: or you write it to the JavaDoc...

gateleen-hook/pom.xml
hiddenalpha: Not important. Only a matter of convenience for development.
Part of commit message:
> ... - Fix logger configuration for tests to make developers life easier.

Logs in test won't affect released artifact. Therefore why not? Are there any problems with this? How do we debug a test failing only on travis not reproducible locally for example?

Never mind. I'll remove it :)

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
hiddenalpha: Because absent (aka null) is already used to indicate _header-not-present_
Cited from javadoc of ExpiryCheckHandler#getExpireValue(String):
> _returns Expire-after time in seconds_ **or null if header not present**.

Actually non existent header is not an error (or did I miss a specification somewhere?). Therefore a caller can decide to use a default value in this case (See: HookHandler:1051).

Where in contrast a X-Expire-After value eg "-1234" or "Hello World" would definitively be a client Problem and therefore should be answered with a 400 response code (In my opinion at least. No spec found which declares type of X-Expire-After).

If we Cut down this API using null (or absent() (doesn't matter)) for both (not-present AND error), then all callers (especially future ones) will never be able to differ those cases. Therefore we'll be forced to extend our API again, again and again violating Interface-Segregation-Principle [1] (or also MinimalInterface [2]) more and more.

Another (more verbose) variant would be to throw a checked exception like MyCheckedNumberFormatException so we'll force callers to handle that case explicitly.

Never mind. I'll try to replace it with a more limited interface  :-)

[1] https://en.wikipedia.org/wiki/Interface_segregation_principle
[2] https://martinfowler.com/bliki/MinimalInterface.html

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HttpHook.java
hiddenalpha: This seems to be _gateleen public API_. Therefore keep in mind that we introduce a _breaking change_ doing this.

Nevertheless I'll apply this  :-)

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
hiddenalpha: Same as https://github.com/swisspush/gateleen/pull/209#discussion_r178061152

But I'll change it :-)

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
hiddenalpha: Oops. Seems valid. 👍

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/mocks/FastFailHttpServerRequest.java
hiddenalpha: Not exactly. But you're right and this can definitively be improved somewhat :-)

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/mocks/FastFailHttpServerResponse.java
hiddenalpha: Not exactly. But you're right and this can definitively be improved somewhat :-)

lbovet:
I also had a look. I have nothing to add to Marc's comment. They are valid and this will mostly lead to more homogeneous code. Thanks for all this work!

mcweba:
@hiddenalpha can you notify me when the changes (2nd run) are ready to review.

hiddenalpha:
Ready to review "3rd run".

Issue #226, opened 2018-06-13
DeltaHandler: delta values expire after 20 days


oli-h:
DeltaHandler has a surprising behavior: delta values (per resource) expire after 20 days if no "x-expire-after" header is given in the PUT request.

At least: This is documented (https://github.com/swisspush/gateleen/blob/v1.1.29/gateleen-delta/README_delta.md) - but the documentation still doesn't explain the reason for this magic "20 days" limit.

Finally this leads to the (documented) situation that a resources looses its delta-information and the required optimization when mirroring a whole subtree (with ExpansionHandler) doesn't work any more. We finally always receive a whole subtree, leading to high bandwidth and high system load.

Propose to remove this magic "20 days" and have a default of "do not expire" - finally the same behavior as vertx-rest-storage for PUT resources

oli-h:
Solved with https://github.com/swisspush/gateleen/pull/227

PR #227, opened 2018-06-13
If no "x-expire-after" is given but "x-delta: auto", then DeltaHandler still added a 20-day-time-to-live

oli-h:
...to the delta values in Redis.
This is fixed here - no "x-expire-after" means "no expire" - for the resource itself but now also for its delta value

solves #226

### DIFF
--- a/gateleen-delta/pom.xml
+++ b/gateleen-delta/pom.xml
@@ -18,5 +18,13 @@
             <artifactId>gateleen-routing</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- TEST dependencies -->
+        <dependency>
+            <groupId>org.swisspush.gateleen</groupId>
+            <artifactId>gateleen-testhelper</artifactId>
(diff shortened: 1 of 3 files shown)


oli-h:
We now have a Unit-Test for the two cases _with_ and _without_ header `x-expire-after`

gateleen-delta/pom.xml
```diff
@@ -18,5 +18,13 @@
             <artifactId>gateleen-routing</artifactId>
             <version>${project.version}</version>
         </dependency>
+
+        <!-- TEST dependencies -->
+        <dependency>
```
mcweba: Is not really used or do I miss something?

oli-h: We get JUnit and Mockito with this dependency. It's `<scope>test</scope>`.
So I don't understand your question?

mcweba: You're right I just saw the AbstractLuaScriptTest class

PR #239, opened 2018-07-10
[#238] Enable route hooks to set a connectionPoolSize

hiddenalpha:
- Defined new hook property 'connectionPoolSize'
- Using default connectionPoolSize of 50 (was 5000 before)

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
@@ -28,7 +28,6 @@ public class Route {
     /** How long to let the http clients live before closing them after a update on the forwarder */
     private static final int GRACE_PERIOD = 30000;
     private static final int CLIENT_DEFAULT_TIMEOUT_SEC = 30;
-    private static final int CLIENT_DEFAULT_POOL_SIZE = 5000;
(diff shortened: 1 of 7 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
```diff
@@ -157,7 +157,19 @@ private void createHttpClient() {
...
         else {
-            HttpClientOptions options = new HttpClientOptions().setDefaultHost(rule.getHost()).setDefaultPort(rule.getPort()).setMaxPoolSize(5000).setKeepAlive(true).setPipelining(false);
+            Integer connectionPoolSize = httpHook.getConnectionPoolSize();
+            if (connectionPoolSize == null) {
+                connectionPoolSize = HttpHook.CONNECTION_POOL_SIZE_DEFAULT_VALUE;
+                LOG.debug("No connectionPoolSize specified for route '{}' using default of {}.", rule.getRuleIdentifier(), connectionPoolSize);
+            } else {
+                LOG.debug("Using connectionPoolSize of {} for route '{}'.", connectionPoolSize, rule.getRuleIdentifier());
+            }
+            HttpClientOptions options = new HttpClientOptions()
```
oli-h: Reuse code from gateleen-routing

PR #242, opened 2018-08-06
240: Add error handling for crashing delta handler

hiddenalpha:
- Attach a general error handler in LocalHttpServerResponse to log such
  issues
- Handle undocumented DecodeException by throwing a corresponding
  ResourceCollectionException

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpServerResponse.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpServerResponse.java
@@ -1,11 +1,16 @@
 package org.swisspush.gateleen.core.http;
 
 import io.vertx.codegen.annotations.Nullable;
-import io.vertx.core.http.*;
-import org.swisspush.gateleen.core.util.StatusCode;
-import io.vertx.core.*;
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
+import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.*;
 import io.vertx.core.net.NetSocket;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.swisspush.gateleen.core.util.StatusCode;
 
 import java.util.List;
 
@@ -16,6 +21,7 @@ import java.util.List;
  */
 public class LocalHttpServerResponse extends BufferBridge implements HttpServerResponse {
 
+    private static final Logger logger = LoggerFactory.getLogger(LocalHttpServerResponse.class);
     private int statusCode;
     private String statusMessage;
     private static final String EMPTY = "";
@@ -130,6 +136,10 @@ public class LocalHttpServerResponse extends BufferBridge implements HttpServerR
 
     public LocalHttpServerResponse(Vertx vertx, Handler<HttpClientResponse> responseHandler) {
         super(vertx);
+        // Attach most simple possible exception handler to base.
+        setExceptionHandler(thr -> {
+            logger.error("Processing of response failed.", thr);
+        });
         this.responseHandler = responseHandler;
     }
 
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
@@ -286,15 +286,17 @@ public final class ExpansionDeltaUtil {
         try {
             obj = new JsonObject(dataAsString);
         } catch (DecodeException e) {
-            final int MAX_CHARS_TO_PRINT = 4096;
-            final String msgHead = "Failed to decode data as JSON.";
-            final String msgTail = "First " + MAX_CHARS_TO_PRINT + " characters were:\n" +
-                    // Append first n characters from malformed data
-                    dataAsString.substring(0, Math.min(MAX_CHARS_TO_PRINT, dataAsString.length())) + "\n";
-            // Log as INFO only because this situation is not our fault and we can continue
-            // our service with no problems.
-            log.info(msgHead, e);
-            throw new ResourceCollectionException(msgHead + "\n\n" + msgTail, StatusCode.BAD_GATEWAY, e);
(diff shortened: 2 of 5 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpServerResponse.java
```diff
@@ -130,6 +136,10 @@ public HttpClientResponse exceptionHandler(Handler<Throwable> handler) {
 
     public LocalHttpServerResponse(Vertx vertx, Handler<HttpClientResponse> responseHandler) {
         super(vertx);
+        // Attach most simple possible exception handler to base.
+        setExceptionHandler(thr -> {
```
oli-h: `thr` is a very uncommon name here. Consider to use `ex` or `e` - even when it's Thowable

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
```diff
@@ -277,7 +280,23 @@ public static CollectionResourceContainer verifyCollectionResponse(String target
...
         }
 
-        JsonObject obj = new JsonObject(data.toString("UTF-8"));
+        // Parse payload
+        final String dataAsString = data.toString(UTF_8);
+        final JsonObject obj;
+        try {
+            obj = new JsonObject(dataAsString);
+        } catch (DecodeException e) {
+            final int MAX_CHARS_TO_PRINT = 4096;
```
oli-h: Why limit?

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
```diff
@@ -277,7 +280,23 @@ public static CollectionResourceContainer verifyCollectionResponse(String target
...
+            final String msgHead = "Failed to decode data as JSON.";
+            final String msgTail = "First " + MAX_CHARS_TO_PRINT + " characters were:\n" +
+                    // Append first n characters from malformed data
+                    dataAsString.substring(0, Math.min(MAX_CHARS_TO_PRINT, dataAsString.length())) + "\n";
+            // Log as INFO only because this situation is not our fault and we can continue
+            // our service with no problems.
+            log.info(msgHead, e);
+            throw new ResourceCollectionException(msgHead + "\n\n" + msgTail, StatusCode.BAD_GATEWAY, e);
```
oli-h: Do we really end the response in this case?
A not-ended response is the original issue (i.e. client is waiting forever for the response).
Can you please explain why this new Exception Handler is enough?

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpServerResponse.java
hiddenalpha: Would name `throwable` (fully written) also be an option? I would prefer that because:

1. By fact an `Exception is a  Throwable` but a `Throwable is NOT an Exception`. Therefore name `e`  would be a lie.

2. This would feel like:

```
interface Pet {}
class Dog implements Pet {}
class Cat implements Pet {}

class EquivalentOfOurErrorHandler {
    void thisIsLikeOurErrorhandler(Pet cat){
        // Takes a 'Pet' but names the variable 'cat' -> what?
        // Hmm Essentially it would be the same if we rename 'thr' to 'ex' in our case.
        // IMHO this sounds somewhat silly. Doesn't it? :)
    }
}
```

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
hiddenalpha: Hmm. Some instinct only.

Should we instead send e.g. 250MB to a client only to tell him, that the response is useless? (Status is 502 and we send some garbage at beginning anyway)

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
hiddenalpha: Yes this suffice :)

Because down the stack in [DeltaHandler.java:255](https://github.com/swisspush/gateleen/blob/v1.1.34/gateleen-delta/src/main/java/org/swisspush/gateleen/delta/DeltaHandler.java#L255) where we do the call, we handle exactly this exception and 'end' the response at [DeltaHandler.java:295](https://github.com/swisspush/gateleen/blob/v1.1.34/gateleen-delta/src/main/java/org/swisspush/gateleen/delta/DeltaHandler.java#L295).

You can observe, that this only adapts error handling like other cases in same method e.g. [ExpansionDeltaUtil.java:240](https://github.com/swisspush/gateleen/pull/242/files/720ff27433515a76553df3926977eb2407f857a8#diff-7a73c7cbdcc00d2dac0ade9446aa3f85L277)

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpServerResponse.java
oli-h: Ok then, you're right. Let's keep `thr` - this is not a show-stopper

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
oli-h: Agree. Some limit makes sense.
Though I have the feeling that 4096 is a bit small. Often Stacktraces themselves are 1-2 kBytes in size.
Suggest to increase limit to 64k (65536) - and can be provide the URL from where we got the non-JSON-parseable response?

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/ExpansionDeltaUtil.java
oli-h: Cool. Thx

hiddenalpha:
Build for d2789a3decae5ed4b67607adb3c74f8897a434d8 "_Apply suggested changes (1st run)_" running.
Please review when it's completed :)

PR #253, opened 2018-11-19
[#249] Enable QueueProcessor to handle other HTTP methods

### LINKED ISSUE #249: QueueProcessor cannot serve Listener-Hooks for PATCH, HEAD, etc
From client side we can set a Listener-Hook for PATCH, HEAD etc. But later QueueProcessor can't serve it.

Eg if we send a PATCH request, head of queue will be a PATCH message (as expected of course). But badly, QueueProcessor fails to handle that PATCH as he only supports GET, PUT, POST and DELETE

See [QueueProcessor.java:69](https://github.com/swisspush/gateleen/blob/v1.1.33/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java#L69) calling [ctor of HttpRequest](https://github.com/swisspush/gateleen/blob/v1.1.33/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L54) which then only accept whitelisted methods listed in [switch at HttpRequest.java:60](https://github.com/swisspush/gateleen/blob/v1.1.33/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L60-L68).

Finally, the queue is blocked with error message like:

    ERROR QueueProcessor - Could not build request: {"queue":"listener-hook-http+reproducer+tmp","payload":"{\"method\":\"PATCH\",\"uri\":\"/my-web-ctxt/server/hooks/v1/listeners/http/foo/bar\",\"headers\":[  /*Headers skipped here*/  ],\"payload\":\"\",\"queueTimestamp\":1234567897750}"} error is Request method must be one of GET, PUT, POST or DELETE

Even the two expiries (x-expire-after and x-queue-expire-after, skipped in example log) don't work as they depend of a successful decoded payload to a HttpMessage object.

hiddenalpha:
- Accept only whitelisted HTTP methods on listener registration which
  namely are: GET, HEAD, PUT, POST, DELETE, OPTIONS and PATCH.
- Respond with a HTTP 400 in case a not whitelisted method got used.
- Document possible exceptions on public API.
- Drop enqueueing requests with non-processable HTTP methods and log
  that attempt.

Fixes #249 .

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
@@ -50,6 +50,11 @@ public class HttpRequest {
 
     /**
      * @param object object
+     * @throws IllegalArgumentException
+     *      When passed in JSON is not in expected format.
+     * @throws ClassCastException
+     *      IMO, this should be a {@link IllegalArgumentException}. Because this
+     *      happens when impl tries to cast something we got from passed in JSON.
      */
     public HttpRequest(JsonObject object) {
         this.method = HttpMethod.valueOf(object.getString("method"));
@@ -58,13 +63,22 @@ public class HttpRequest {
             throw new IllegalArgumentException("Request fields 'uri' and 'method' must be set");
         }
         switch (method) {
+        // We accept those methods:
         case GET:
+        case HEAD:
         case PUT:
         case POST:
         case DELETE:
+        case OPTIONS:
+        case PATCH:
             break;
         default:
-            throw new IllegalArgumentException("Request method must be one of GET, PUT, POST or DELETE");
+            // This (default branch) gets reached when:
+            // 1. Someone is using a (for us) unknown http method (eg someone added a new
+            //    one in the enum since this here was written).
+            // 2. We explicitly forbid CONNECT, TRACE and OTHER.
+            //    See "https://github.com/swisspush/gateleen/issues/249".
+            throw new IllegalArgumentException("Request method must be one of GET, HEAD, PUT, POST, DELETE, OPTIONS or PATCH");
         }
         JsonArray headersArray = object.getJsonArray("headers");
         if (headersArray != null) {
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
@@ -1,10 +1,10 @@
 package org.swisspush.gateleen.core.http;
 
-import org.swisspush.gateleen.core.json.JsonMultiMap;
 import io.vertx.core.MultiMap;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
+import org.swisspush.gateleen.core.json.JsonMultiMap;
 
 import java.util.Arrays;
 
@@ -63,14 +63,14 @@ public class HttpRequest {
             throw new IllegalArgumentException("Request fields 'uri' and 'method' must be set");
         }
         switch (method) {
-        // We accept those methods:
+            // We accept those methods:
         case GET:
-        case HEAD:
+            case HEAD:
         case PUT:
         case POST:
         case DELETE:
-        case OPTIONS:
-        case PATCH:
+            case OPTIONS:
+            case PATCH:
             break;
         default:
             // This (default branch) gets reached when:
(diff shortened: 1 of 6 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
```diff
@@ -50,6 +50,11 @@ public HttpRequest(HttpMethod method, String uri, MultiMap headers, byte[] paylo
 
     /**
      * @param object object
+     * @throws IllegalArgumentException
+     *      When passed in JSON is not in expected format.
+     * @throws ClassCastException
```
mcweba: Does it throw both exceptions now? If not change the javadoc

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1061,6 +1062,33 @@ private void handleListenerRegistration(final HttpServerRequest request) {
...
+            hook = new JsonObject(hookData);
+        } catch (DecodeException e) {
+            log.error("Cannot decode JSON", e);
+            badRequest(request, "Cannot decode JSON", e.getMessage());
+            return true;
+        }
+        final JsonArray methods = hook.getJsonArray("methods");
+        for (Object method : methods) {
+            if( !QueueProcessor.httpMethodIsQueueable( HttpMethod.valueOf((String)method) ) ){
+                final String msg = "Listener registration request tries to hook for forbidden '"+method+"' method.";
```
mcweba: I would not use the term "forbidden" in this log message. Better use "not allowed" since forbidden implies that the client has insufficient permissions to make the request which is not true.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1061,6 +1062,33 @@ private void handleListenerRegistration(final HttpServerRequest request) {
...
+                log.error( msg );
+                badRequest(request, "Bad Request", msg+"\n" );
```
mcweba: Do not respond this requests with "400 Bad Request". Use "405 Method not Allowed"

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
hiddenalpha: There's nothing new with those exceptions. They were already there before my change. I only documented them because that was missing and would help a lot to write more reliable code. Here some links pointing to previous release:

**IllegalArgumentException**:
- Thrown at lines [58](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L58), [67](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L67) and [73](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L73).

**ClassCastException**:
- Thrown at lines [55](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L55), [56](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L56), [69](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L69) and [73](https://github.com/swisspush/gateleen/blob/develop/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java#L73).

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
hiddenalpha: Ok. Will do 👍

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
mcweba: Good point. I did not thought about this. Therefore it's OK to respond with a 400

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -479,11 +489,11 @@ public void listenerRegistration_rejectNotWhitelistedHttpMethods(TestContext tes
             }
 
             // Trigger
-            hookHandler.handle( request );
+            hookHandler.handle(request);
 
             { // Assert request got rejected.
-                testContext.assertTrue( statusCodePtr[0] >= 400 );
-                testContext.assertTrue( statusCodePtr[0] <= 499 );
+                testContext.assertTrue(statusCodePtr[0] >= 400);
```
mcweba: Why not checking that status code is 400 when requests with these methods are made?

hiddenalpha: Hmm. Yes would be an option of course. Didn't think about that. Will do 👍

Issue #255, opened 2018-11-26
Forwarder (gateleen-routing) sets "transfer-encoding: chunked" also for responses without body


oli-h:
Problematic code part:
https://github.com/swisspush/gateleen/blob/f66bf9e73bbd06e064e50916dec3ca0aaa7bb17b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java#L367-L369

This basically works fine, i.e. if we have no content-length header from our upstream server then we use chunked for response to our client. Problem:
- some http-Responses don't contain a body (not even a body with length=0). [According to http-Specification](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.3):
> All 1xx (informational), 204 (no content), and 304 (not modified) responses MUST NOT include a message-body
- When the upstream now sends a correct http-response to us (i.e. no Content-Length, no Transfer-encoding, no body bytes) then our current implementation still sets "chunked" 
_(which leads to a "Transfer-encoding: chunked" header and (at least) the final '0' chunk indication in the response stream)_
- Some clients (e.g. Browser Chrome, JDK's internal Http-Client) are not amused about this (e.g. response-body for 304 not modified or PUT requests) and closes the connection after the http-cycle --> we loose keep-alive on this connection.

Propose to change the behaviour:
- We already forward (nearly all) response-headers, including content-length, transfer-encoding (or probably none of them) - we now just have to say Vert.x 'chunked' if our upstream said 'chunked'. Therefore change the code to:
```
if (req.response().headers().contains(HttpHeaders.TRANSFER_ENCODING, "chunked", true)) {
   req.response().setChunked(true);
}
```

Issue #254, opened 2018-11-22
Forwarder (gateleen-routing) can leave HttpConnection to upstream server in 'paused()' mode


oli-h:
Gateleen's Forwarder in fact is a Http-Proxy.
There are situations when a client (e.g. Webbrowser) closes a TCP-Stream (TCP-FIN) while the transfer of a http-Response-body is still in progess (e.g. Video-Stream).

In special cases (upstream-server delivers faster than we can deliver to Webbrowser), Gateleen's Forwarder pauses the upstream with help of Vertx's `Pump` (back pressure). If now the WebBrowser TCP-FIN it's connection to us, then the upstream http-cycle never finishes and stays in this 'paused' state forever.

Depending on the "max connections" on this Forwarder we run out of connection for this route, leading to "504 timeouts" for further WebBrowser requests

PR #256, opened 2018-11-27
Bugfix issues 254 255 stablize proxy behaviour

oli-h:
Two main points are addressed:
- #255 correct http-Responses for Gateleen's LocalClient (concerning header 'Content-Length' and 'chunked'-Transfer).
_This is relevant for Gateleen's forwarder so it's is not required to fallback to 'chunked' in absence of a content-length header - and this 'chunked' is problematic for Clients/User-Agents in cases where http-Spec does not allow any body (e.g. "304 not modified")_
- #254 resume (i.e. un-pause) an upstream-response in case of errornous situations
_This is relevant as otherwise there are cases where such 'paused' streams are left in starvation and can't be used any more_


gateleen-core/src/main/java/org/swisspush/gateleen/core/http/LocalHttpClientRequest.java
```diff
@@ -473,6 +475,113 @@ public LocalHttpClientRequest(HttpMethod method, String uri, Vertx vertx, Handle
         this.uri = uri;
         this.routingContextHandler = routingContextHandler;
         this.response = response;
+        // this is a mock-connection object which silently ignores (nearly) HTTP/1.1 relevant method calls and never throws exceptions
+        // we need this to be able to set appropriate exception- and close-handlers in e.g. Gateleen's Forwarder (gateleen-routing)
+        this.connection = new HttpConnection() {
```
hiddenalpha: I suggest to split this up as follows:

1) Introduce a class with its single purpose to throw no matter which method gets called:

```
class FailFastHttpConnection implements HttpConnection {
    
    private static final String msg = "This (mocked) connection doesn't support that call.";

    // Implement ALL methods to simply throw an exception.
    @Override
    public HttpConnection goAway(long errorCode, int lastStreamId, Buffer debugData) {
        throw new UnsupportedOperationException( msg );
    }
}
```

2) Then to create our instance:

```
    this.connection = new FailFastHttpConnection(){
        // Now here we only need to implement special cases because all other
        // methods we don't care about already throw in our super class.
        @Override
        public HttpConnection closeHandler(Handler<Void> handler) {
            return this;
        }
    };
```

This IMO would be:

1) Much simpler to read and understand.
2) Our extra class can simply be reused.
3) No distracting code bloat in our logic.

Issue #268, opened 2019-03-12
Add proxy options to routing rules


mcweba:
The router creates _Forwarder_ instances for each routing rule. Each _Forwarder_ has its own _HttpClient_ instance which can have their own proxy options.

The current implementation of the routing rules does not provide the possiblity to configure the proxy options for a single routing rule.

## ToDo
Extend the routing rules to provide proxy options which can be mapped to [https://vertx.io/docs/apidocs/io/vertx/core/net/ProxyOptions.html](https://vertx.io/docs/apidocs/io/vertx/core/net/ProxyOptions.html)

Example:
```json
{
	"/some/url/(.*)": {
		"description": "routing rule having a proxy config",
		"url": "https://somehost:1234/another/url/$1",
		"proxyOptions": {
			"type": "http",
			"host": "somehost",
			"port": 3128
		}
	}
}
```
The _proxyOptions_ should also be validated with the json schema for routing rules

mcweba:
Released in [v1.1.42](https://github.com/swisspush/gateleen/releases/tag/v1.1.42)

PR #269, opened 2019-03-12
#268 added proxy options to routing rules and hook config

mcweba:


gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
```diff
@@ -132,6 +136,32 @@
 			],
 			"additionalProperties": false
 		},
+		"ProxyOptions": {
+			"properties": {
+				"type": {
+                    "type": "string",
```
oli-h: indentation

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
```diff
@@ -132,6 +136,32 @@
...
+                        "SOCKS4",
+                        "SOCKS5"
+                    ]
+				},
+				"host": {
+					"type": "string"
+				},
+				"port": {
+					"type": "integer",
+					"minimum": 0
```
oli-h: maximum 65535?

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
```diff
@@ -132,6 +136,32 @@
...
+				"type": {
+                    "type": "string",
+                    "enum": [
+                        "HTTP",
+                        "SOCKS4",
+                        "SOCKS5"
+                    ]
+				},
+				"host": {
+					"type": "string"
```
oli-h: can we specify in this schema that this is a **required** attribute? (And also port - but not username/password)?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1305,11 +1308,28 @@ private void registerListener(Buffer buffer) {
...
 
+    /**
+     * Extract proxyOptions attribute from jsonHook and create a
+     * appropriate property in the hook object.
+     *
+     * This is the same concept as in gateleen-routing:
+     * {@link org.swisspush.gateleen.routing.RuleFactory#setProxyOptions(Rule, JsonObject)}}
+     *
+     * @param jsonHook the json hook
+     * @param hook the hook object
```
hiddenalpha: IMHO those _@param_ descriptions are useless. They only repeat the parameter name. IMHO thats no additional value.

Is there any sense to keep such doc?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RuleFactory.java
```diff
@@ -136,6 +138,13 @@ private void setTranslateStatus(Rule ruleObj, JsonObject rule) {
         }
     }
 
+    private void setProxyOptions(Rule ruleObj, JsonObject rule) {
+        JsonObject proxyOptions = rule.getJsonObject("proxyOptions");
+        if(proxyOptions != null){
```
oli-h: code format

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
```diff
@@ -198,6 +202,11 @@ public HttpClientOptions buildHttpClientOptions() {
         if ("https".equals(getScheme())) {
             options.setSsl(true).setVerifyHost(false).setTrustAll(true);
         }
+
+        if(getProxyOptions() != null){
```
oli-h: code format

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1305,11 +1308,28 @@ private void registerListener(Buffer buffer) {
...
+            hook.setProxyOptions(new ProxyOptions(proxyOptions));
+        }
+    }
+
     /**
      * Extract staticHeaders attribute from jsonHook and create a
      * appropriate list in the hook object.
      *
-     * This is the same concept as in gateleen-rooting:
+     * This is the same concept as in gateleen-routing:
```
hiddenalpha: Nice catch ;)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
```diff
@@ -182,9 +184,11 @@ public String getStorage() {
         return storage;
     }
 
-    public void setStorage(String storage) {
-        this.storage = storage;
-    }
+    public void setStorage(String storage) { this.storage = storage; }
+
+    public ProxyOptions getProxyOptions() { return proxyOptions; }
```
oli-h: code format - this is just a one-liner. Weird.
(although IDEs typically present such one-liner-methods in this compact form it should not be a one-liner in the Sourcecode-File itself)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
```diff
@@ -182,9 +184,11 @@ public String getStorage() {
         return storage;
     }
 
-    public void setStorage(String storage) {
-        this.storage = storage;
-    }
+    public void setStorage(String storage) { this.storage = storage; }
```
oli-h: code format - this is just a one-liner. Weird.
(although IDEs typically present such one-liner-methods in this compact form it should not be a one-liner in the Sourcecode-File itself)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
```diff
@@ -344,7 +342,9 @@ private void createForwarders(List<Rule> rules, io.vertx.ext.web.Router newRoute
             } else if (rule.getScheme().equals("local")) {
                 forwarder = new Forwarder(vertx, selfClient, rule, this.storage, loggingResourceManager, monitoringHandler, userProfileUri);
             } else {
+                HttpClient client = vertx.createHttpClient( rule.buildHttpClientOptions() );
```
oli-h: interesting finding - we really created HttpClients for "local" routes and didn't recognized this before?

gateleen-hook/src/main/resources/gateleen_hooking_schema_hook
```diff
@@ -131,6 +135,32 @@
...
+				},
+				"username": {
+					"type": "string"
+				},
+				"password": {
+					"type": "string"
+				}
+			},
+			"additionalProperties": false
+		}
```
hiddenalpha: **Bad indentation:** Some of those lines are indented by tabs whereas rest of file uses 4 spaces.

gateleen-hook/src/main/resources/gateleen_hooking_schema_hook
```diff
@@ -131,6 +135,32 @@
...
+				"type": {
+                    "type": "string",
+                    "enum": [
+                        "HTTP",
+                        "SOCKS4",
+                        "SOCKS5"
+                    ]
+				},
+				"host": {
+					"type": "string"
```
oli-h: same comments as for the routing-rule-schema.
It would be good to specify "required" attributes

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
oli-h: *smile* route to root

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1305,11 +1308,28 @@ private void registerListener(Buffer buffer) {
...
+     *
+     * This is the same concept as in gateleen-routing:
+     * {@link org.swisspush.gateleen.routing.RuleFactory#setProxyOptions(Rule, JsonObject)}}
+     *
+     * @param jsonHook the json hook
+     * @param hook the hook object
+     */
+    private void extractAndAddProxyOptionsToHook(final JsonObject jsonHook, final HttpHook hook) {
+        JsonObject proxyOptions = jsonHook.getJsonObject("proxyOptions");
+        if(proxyOptions != null){
```
oli-h: code format

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
```diff
@@ -132,6 +136,32 @@
...
+				},
+				"username": {
+					"type": "string"
+				},
+				"password": {
+					"type": "string"
+				}
+			},
+			"additionalProperties": false
+		},
```
hiddenalpha: **Bad indentation:** Some of those lines are indent by _4 spaces_ where rest of file uses _1 tab_.

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
hiddenalpha: AFAIK we could with:
```
"properties": { ... },
"additionalProperties": false,
"required": [ "host" , "port" ]
```

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
ljucam: code format

hiddenalpha: I found this during #239 (See [diff](https://github.com/swisspush/gateleen/pull/239/files#diff-8fa3f5fa721d37d2dad6ccd9edde2d7cR333)) (do you remind?). Back then I didn't change that because I didn't know about possible side effects and added that TODO instead.

hiddenalpha: :( Sad. I'm in love with this style.

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
mcweba: of course we can. The question is whether we want or not? none of these properties are required since there's a default value (at least for type, host and port)

mcweba: I will add _host_ and _port_ as required since when someone wants to use the _proxyOptions_ property he should define something. I think the _type_ is not mandatory, because the default value is _HTTP_ which probably will be used in most cases.

hiddenalpha: **host:** What default would you think for then? IMHO a default would be _no_ proxy. But then, whole _proxy_ object would make any sense.

**port:** Which defaults would be intuitive? IMHO that would depend on _type_. But even then it seems there are tons of different "defaults". Eg search for "proxy" on [List of TCP and UDP port numbers](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers).

PR #279, opened 2019-06-21
Issue278 requestheaders in json transform

mcweba:


gateleen-core/src/main/java/org/swisspush/gateleen/core/json/transform/JoltSpecBuilder.java
```diff
@@ -13,20 +13,28 @@
...
      * If an error occurs during the creation of the spec, a {@link JoltSpecException} is thrown.
      *
      * @param jsonSpec the string representation of the spec
+     * @param withMetadata does the json input to use the spec against has meta data
      * @return returns a {@link JoltSpec}
-     * @throws JoltSpecException when an error occured during spec parsing
+     * @throws JoltSpecException when an error occurred during spec parsing
      */
-    public static JoltSpec buildSpec(String jsonSpec) throws JoltSpecException {
+    public static JoltSpec buildSpecWithMetadata(String jsonSpec, boolean withMetadata) throws JoltSpecException {
```
hiddenalpha: Suggest to drop `WithMetadata` from method name.

> // Example call
> buildSpecWithMetadata( foo , false );

IMHO that call sounds somewhat counter-intuitive.

oli-h:
Didn't follow the issue in details - just a hint: Consider zo use the "header functions" (as we have in Hooks and Routes)

PR #286, opened 2019-08-22
Extend ACL with RoleMapping (Issue #285)

### LINKED ISSUE #285: Extend ACL with RoleMapping
Often we have different ACL roles which have equal setup in Gateleen because those roles are used within backend applications only for further distinguishing roles.  In order to reduce configuration efforts and redundancy, Roles should be treated as well with Wildcard in addition in Gateleen Security module (acl):

Example:
z-paisa-fis-admin
z-paisa-fis-user
z-paisa-fis-manager

We would like to add only one ACL for  z-paisa-fis

Reference:  SDCISA-409

Kusig:
fixes #285

Introduction of RoleMapper for ACL role mapping
See README_security.md


gateleen-security/src/test/java/org/swisspush/gateleen/security/authorization/AuthorizerTest.java
```diff
@@ -240,16 +244,67 @@ public void testHandleNotIsAuthorized(TestContext context) {
...
+        AuthorizerRequest req = new AuthorizerRequest(HttpMethod.PUT, requestUri, headers, response);
+
+        long start = System.currentTimeMillis();
+        System.out.println("Start: " + start);
+        int i;
+        for (i=0; i<100000; i++) {
+            authorizer.authorize(req);
+        }
+        long end = System.currentTimeMillis();
+        System.out.println("End:" + end + " Loops:" + i + " Time=" + (end-start));
```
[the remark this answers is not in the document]
Kusig: Removed.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/AclFactory.java
```diff
@@ -72,4 +76,28 @@ private void checkPropertiesValid(String path, JsonArray methods, String id) thr
...
+
+        List<RoleMapperHolder> result = new ArrayList<>();
+        JsonObject mapItems = new JsonObject(buffer.toString("UTF-8"));
+        JsonArray mappers = mapItems.getJsonArray("rolemappers");
+        for (Object obj : mappers) {
+            JsonObject mapper = new JsonObject((obj.toString()));
+            String pattern = mapper.getString("pattern");
+            String role = mapper.getString("role");
+            Boolean keepOriginal = mapper.getBoolean("keeporiginal");
+            if (pattern != null && !pattern.isEmpty() && role != null && !role.isEmpty() && keepOriginal!=null) {
```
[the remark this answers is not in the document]
Kusig: Looking at isEmpty() shows that it returns TRUE as well if the string is null, dangerous construct ...

However, in my case the result is the same and I will change it therefore.

gateleen-security/src/test/java/org/swisspush/gateleen/security/authorization/AclFactoryTest.java
```diff
@@ -30,15 +30,16 @@
...
+    private String mapperSchema = ResourcesUtils.loadResource("gateleen_security_schema_rolemapper", true);
 
     private final String INVALID_JSON = ResourcesUtils.loadResource("testresource_invalid_json", true);
     private final String VALID_ACL_RESOURCE = ResourcesUtils.loadResource("testresource_valid_acl_resource", true);
     private final String ADDITIONAL_PROP_ACL_RESOURCE = ResourcesUtils.loadResource("testresource_additionalproperties_acl_resource", true);
 
     @Before
     public void setUp(){
-        aclFactory = new AclFactory(schedulersSchema);
+        aclFactory = new AclFactory(aclSchema,mapperSchema);
```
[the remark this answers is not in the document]
Kusig: Ok, catched lazy boss :-(

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -271,6 +383,47 @@ private boolean isAuthorized(Set<String> roles, HttpServerRequest request) {
...
     }
 
+    /**
+     * Maps the received roles from http header according the rolemapper rules and return the set of
+     * mapped roles including the initial list of roles.
+     *
+     * @param roles The roles to be mapped and enrichted according to the rolemapper object
+     * @return The resulting list of initial plus mapped roles
+     */
+    private Set<String> mapRoles(Set<String> roles) {
```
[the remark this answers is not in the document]
Kusig: Ok, this one was unfair :-)  Resulting in a total refactoring of the whole stuff to make it more modular ...  
New big push will come ...

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -221,13 +239,70 @@ private void handleAclUriRequest(final HttpServerRequest request, Future<Boolean
         }
     }
 
+    private void handleRoleMapperUriRequest(final HttpServerRequest request, Future<Boolean> future) {
```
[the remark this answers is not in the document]
Kusig: As well modularised and made it Interfacing from a base Configuration BaseClass in order to make these duplications go.
In fact this is probably something we should even make available this was Gateleen wide, I suspect this is used like this over and over again ....

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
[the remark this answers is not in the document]
mcweba: Welcome to the developer's life :-)

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -67,47 +63,42 @@ public Authorizer(Vertx vertx, final ResourceStorage storage, String securityRoo
         this.aclRoot = securityRoot + aclKey + "/";
         this.aclUriPattern = new PatternHolder(Pattern.compile("^" + aclRoot + "(?<role>.+)$"));
         this.userUriPattern = Pattern.compile(securityRoot + "user(\\?.*)?");
+        this.roleMapperUriPattern = new PatternHolder(Pattern.compile("^" + securityRoot + "/rolemapper"));
         this.roleExtractor = new RoleExtractor(rolePattern);
-
-        this.aclSchema = ResourcesUtils.loadResource("gateleen_security_schema_acl", true);
-        this.aclFactory = new AclFactory(aclSchema);
+        this.roleMapper = new RoleMapper(storage,securityRoot);
```
oli-h: code format - need space after comma.
This is not a blocker for me - still, if you touch the code again, please format to avoid future diffs

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -67,47 +63,42 @@ public Authorizer(Vertx vertx, final ResourceStorage storage, String securityRoo
...
+        this.roleAuthorizer = new RoleAuthorizer(storage,securityRoot,rolePattern,roleMapper);
```
oli-h: dito: code format - need space after comma

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperHolder.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    private String role;
+
+    /**
+     * Defines if the originalRole must be kept or not in the list of roles to apply against the ACL
+     */
+    private Boolean keepOriginal;
+
+    RoleMapperHolder(Pattern pattern, String role, Boolean keepOriginal) {
+        super();
```
oli-h: unusual to explicit call non-arg super-constructor....

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperHolder.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void setRole(String role) {
+        this.role = role;
+    }
+
+    public Boolean getKeepOriginal() {
+        return keepOriginal;
+    }
+
+    public void setKeepOriginal(Boolean keepOriginal) {
```
oli-h: This is a *required* attribute according to the JSON-schema - right?
Then you should use `boolean` and not `java.lang.Boolean` to express a non-null value

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperHolder.java
```diff
@@ -0,0 +1,56 @@
...
+        this.pattern = pattern;
+        this.role = role;
+        this.keepOriginal = keepOriginal;
+    }
+
+    public Pattern getPattern() {
+        return pattern;
+    }
+
+    public void setPattern(Pattern pattern) {
```
oli-h: propose to remove all setter-methods and make the three attributes `final`.
This expresses the "read-only" character of this object (as it is loaded/initialized only once)

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperHolder.java
```diff
@@ -0,0 +1,56 @@
...
+     * The resulting role if the given regex pattern matches
+     */
+    private String role;
+
+    /**
+     * Defines if the originalRole must be kept or not in the list of roles to apply against the ACL
+     */
+    private Boolean keepOriginal;
+
+    RoleMapperHolder(Pattern pattern, String role, Boolean keepOriginal) {
```
oli-h: Propose to add JavaDoc and explain that all three parameters must be !=null

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
```diff
@@ -0,0 +1,54 @@
...
+        ValidationResult validationResult = Validator.validateStatic(buffer, mapperSchema, log);
+        if (!validationResult.isSuccess()) {
+           throw new ValidationException(validationResult);
+        }
+
+        List<RoleMapperHolder> result = new ArrayList<>();
+        JsonObject mapItems = new JsonObject(buffer.toString("UTF-8"));
+        JsonArray mappers = mapItems.getJsonArray("mappings");
+        for (Object obj : mappers) {
+            JsonObject mapper = new JsonObject((obj.toString()));
```
oli-h: This is not a *mapper* - it's a "Json-Object"

oli-h: You could directly map to a Java-Object - no need to access single *Json-Attributes by name*

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+
+    public List<RoleMapperHolder> parseRoleMapper(Buffer buffer) throws ValidationException {
+        ValidationResult validationResult = Validator.validateStatic(buffer, mapperSchema, log);
+        if (!validationResult.isSuccess()) {
+           throw new ValidationException(validationResult);
+        }
+
+        List<RoleMapperHolder> result = new ArrayList<>();
+        JsonObject mapItems = new JsonObject(buffer.toString("UTF-8"));
```
oli-h: You can directly consume the bytes to convert to JsonObject - no need to convert to a Java-String.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapper.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private String roleMapper;
+
+    private RoleMapperFactory roleMapperFactory;
+
+    private List<RoleMapperHolder> roleMappers = null;
+
+    public static final Logger log = LoggerFactory.getLogger(RoleMapper.class);
+
+    public RoleMapper( ResourceStorage storage, String securityRoot) {
```
oli-h: code format

oli-h: Please add JavaDoc, espacially to clarify parameter `securityRoot`.
Note that you implicit expect a trailing-slash - but you don't enforce it. So at lease document it

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleAuthorizer.java
```diff
@@ -0,0 +1,217 @@
...
+                    authorized &= (uriRole != null && roles.contains(uriRole));
+                } catch (IllegalArgumentException e) {
+                    // ignore
+                }
+            }
+            if (authorized) {
+                return true;
+            }
+        }
+        return false;
```
oli-h: you could condense code to one single `return authorized` statement

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleAuthorizer.java
```diff
@@ -0,0 +1,217 @@
...
+    public void checkConfigResource(Buffer buffer) throws ValidationException {
+        aclFactory.parseAcl(buffer);
+    }
+
+    @Override
+    public void configUpdate() {
+        storage.get(aclRoot, buffer -> {
+            if (buffer != null) {
+                grantedRoles = new HashMap<>();
+                for (Object roleObject : new JsonObject(buffer.toString("UTF-8")).getJsonArray(aclKey)) {
```
oli-h: No need to convert to a String - you can directly use `buffer` to convert to JsonObject

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleAuthorizer.java
```diff
@@ -0,0 +1,217 @@
...
+     * against the defined RoleMapper if any.
+     *
+     * @param roles The roles used to validate the ACL. Note that the given Roles
+     *              are mapped according to the RoleMapper (if any) before applied against the ACL.
+     * @param request The incoming HTTP Request
+     * @return true if the user is authorized
+     */
+    private boolean isAuthorized(Set<String> roles, HttpServerRequest request) {
+        Set<String> mappedRoles = roleMapper.mapRoles(roles);
+        for (Entry<PatternHolder, Map<String, Set<String>>> entry : grantedRoles.entrySet()) {
```
oli-h: Note that we have a unclear situation during startup:
Loading of the ACLs is async, so `grantedRoles` is filled step by step. If you have requests *while* this filling-up is in progress you can (accidentially) answer such requests with `403 FORBIDDEN` but we probably should answer with `503 Service Unavailable`

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/ConfigurationResource.java
```diff
@@ -0,0 +1,24 @@
+package org.swisspush.gateleen.security.authorization;
+
+import org.swisspush.gateleen.validation.ValidationException;
+import io.vertx.core.buffer.Buffer;
+
+/**
+ * Interface declaring the method to be commonly used for role related resource configuration
+ */
+interface ConfigurationResource {
```
oli-h: consider making it public so other implementations (from other packages) are possible

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -119,20 +110,24 @@ public void authorize(final HttpServerRequest request, final Handler<Void> handl
...
         handleUserUriRequest(request, future);
 
-        if(!future.isComplete()){
-            handleIsAuthorized(request, future);
+        if (!future.isComplete()) {
+            roleAuthorizer.handleIsAuthorized(request, future);
+        }
+
+        if (!future.isComplete()) {
+            handleConfigurationUriRequest(request, future, aclUriPattern, roleAuthorizer);
```
oli-h: `aclUriPattern` and `roleAuthorizer` are instance-variables which you pass to an instance-method.
But the method can access the instance-variables by itself - or do I miss something?

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -67,47 +63,42 @@ public Authorizer(Vertx vertx, final ResourceStorage storage, String securityRoo
...
 
-        if(!future.isComplete()){
-            handleAclUriRequest(request, future);
+        if (!future.isComplete()) {
+            handleConfigurationUriRequest(request, future, aclUriPattern, roleAuthorizer);
         }
 
-        if(!future.isComplete()) {
+        if (!future.isComplete()) {
+            handleConfigurationUriRequest(request, future, roleMapperUriPattern, roleMapper);
```
oli-h: Hmmm.. The `future` will be resolved async (if at all) - but you check 'now' if the `future.isComplete()`.
The whole pattern to work with a future here seems a little bit strange... Why not simply return a boolean which say 'true' or 'false'?

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/ConfigurationResource.java
mcweba: Here I disagree with @oli-h. See my comment in the _Authorizer_ class about using _ConfigurationResourceManager_. I would not further propagate the usage of this _ConfigurationResource_ interface when we already have a solution for this in gateleen.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperHolder.java
Kusig: Could even shown with @NotNull in this case, any problem with that?

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
```diff
@@ -73,52 +63,39 @@ public Authorizer(Vertx vertx, final ResourceStorage storage, String securityRoo
         this.aclRoot = securityRoot + aclKey + "/";
         this.aclUriPattern = new PatternHolder(Pattern.compile("^" + aclRoot + "(?<role>.+)$"));
         this.userUriPattern = Pattern.compile(securityRoot + "user(\\?.*)?");
+        this.roleMapperUriPattern = new PatternHolder(Pattern.compile("^" + securityRoot + "/rolemapper"));
```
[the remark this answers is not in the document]
Kusig: The leading slash istn't necessary in fact. Removed.
Anyhow I propose to make this uri concatenations in a more robust way and introduced a UriBuilder Helper for this.  Used it here initially to check if this is helpful and provoke some more challenging review comments.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
Kusig: Hmm, wasn't sure about the UTF-8 represtation. If I give it the buffer directly, what encoding is it using? I think it is anyhow converted to a String this way as well.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
Kusig: The method is called with different Objects and patterns.  Don't think this is a problem here.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapper.java
Kusig: The method should now be smart enough to concat the right way no matter what kind of slash patterns are provided.
Added JavaDoc

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/ConfigurationResource.java
Kusig: I'll keep it local therefore. Should be once rewritten to use common implementation

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/Authorizer.java
Kusig: Huuuu, this was already like this. Didn't touch it really.
Probably done like this because the caller of the method uses the future for further treatment ..

Kusig: I propose to leave it as is. Would cause again major changes out of scope of the initial task.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleAuthorizer.java
Kusig: As well here this is as it was since ages. If it is worth, please open a new ticket for this. I'll be happy to do this later in a new adventure.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
Kusig: You mean by using Jackson's ObjectMapper?   But I would like to explicitely check the attributes for valid values before creating the RoleMapper object.

gateleen-security/src/test/java/org/swisspush/gateleen/security/authorization/RoleMapperFactoryTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.RunWith;
+import org.swisspush.gateleen.core.util.ResourcesUtils;
+import org.swisspush.gateleen.validation.ValidationException;
+
+import java.util.*;
+
+/**
+ * Tests for the {@link AclFactory} class
```
mcweba: Javadoc should link to _RoleMapperFactory_

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
Kusig: Ok, done as discussed. Validation done in JSON schema properly for non empty values with a Test of course.

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/UriBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Clean the given URI from any multi slash delimiters in between uri segments
+     *
+     * @param uri to be cleaned from possible multi slash delimiters
+     * @return The cleaned uri
+     */
+    public static String cleanUri(String uri) {
+        return uri.replaceAll("(?<!(http:|https:))/+", "/");
```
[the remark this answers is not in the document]
Kusig: Added comment and optimised with precompile PatternMatcher

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.regex.Pattern;
+
+/**
+ * RoleMapperFactory is used to parse RoleMapper resources.
+ */
+public class RoleMapperFactory {
+
+    private String mapperSchema;
+
+    private Logger log = LoggerFactory.getLogger(RoleMapperFactory.class);
```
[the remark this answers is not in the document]
Kusig: Never do C&P ....

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/UriBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+
+
+import java.util.regex.Pattern;
+
+/**
+ * Helper Class for URI manipulations
+ */
+public class UriBuilder {
+
+    private static Pattern uriCleanPattern = Pattern.compile("(?<!(http:|https:))/+");
```
oli-h: *Constants* in Java are typically 1. `static` 2. `final` and 3. named `UPPER_CASE`
This is wide used convention and we should *make it so*

gateleen-security/src/test/resources/testresource_emptyproperties_mapper_resource
```diff
@@ -0,0 +1,8 @@
+{
+  "mappings": [{
+    "pattern":"",
+    "role":"",
+    "keepOriginal":false
+    }
+  ]
+}
```
oli-h: Github mentions a missing final "newline" - probably not an issue

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+        List<RoleMapperHolder> result = new ArrayList<>();
+        Mappings mappings = Json.decodeValue(buffer, Mappings.class);
+        for (Mapping mapping : mappings.mappings) {
+            result.add(new RoleMapperHolder(Pattern.compile(mapping.pattern), mapping.role, mapping.keepOriginal));
+        }
+        return result;
+    }
+
+    private static class Mappings {
```
oli-h: I'm surprised that JSON-Mapping works with `private` classes.
Jackson then probably needs to create "accessor"-Classes which can be less performant.

Independent of this I propose to declare them `public`

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/UriBuilder.java
Kusig: Arghh

gateleen-security/src/test/resources/testresource_emptyproperties_mapper_resource
Kusig: no idea what is missing here.

gateleen-security/src/main/java/org/swisspush/gateleen/security/authorization/RoleMapperFactory.java
Kusig: Tested even with private constructor. Works well. But changed it to public anyway to be on the save side.

gateleen-security/src/test/resources/testresource_emptyproperties_mapper_resource
hiddenalpha: See [why-should-text-files-end-with-a-newline](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

PR #287, opened 2019-08-26
provides dynamic header manipulations for the CopyResourceHandler

gedestroy:
Same as https://github.com/swisspush/gateleen/pull/201 but for the CopyResourceHandler

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/resource/CopyResourceHandler.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/resource/CopyResourceHandler.java
@@ -48,8 +48,12 @@ public class CopyResourceHandler {
             // process task
             request.bodyHandler(buffer -> {
                 CopyTask task = createCopyTask(request, buffer);
-                if (validTask(request, task)) {
+                if (task != null && validTask(request, task)) {
                     performGETRequest(request, task);
+                } else {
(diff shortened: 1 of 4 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/resource/CopyResourceHandler.java
```diff
@@ -175,23 +175,27 @@ private void createResponse(final HttpServerRequest request, final HttpClientRes
...
+        } else {
+            final JsonArray dynamicHeaders = task.getJsonArray("headers");
+            if (dynamicHeaders != null) {
+                headerFunction = HeaderFunctions.parseFromJson(dynamicHeaders);
             }
         }
 
         MultiMap headers = StatusCodeTranslator.getTranslateFreeHeaders(request.headers());
-        headers = HttpHeaderUtil.removeNonForwardHeaders(headers);
+        headerFunction.apply(headers);
```
oli-h: The HeaderFunction returns an `EvalScope` object which contains error information.
Propose to answer the Http-Request with "400 Bad Request".

See https://github.com/swisspush/gateleen/blob/v1.1.46/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java#L197-L205 in Gateleen's Router

gedestroy: OK, right. Will look into this.

Issue #291, opened 2019-09-11
introduce gateleen-kafka module


mcweba:
We want to provide a gateleen-module using the [vertx-kafka-client]( https://vertx.io/docs/vertx-kafka-client/java/). This module should provide a _handler_ class which can be used to forward requests to a kafka topic

mcweba:
Released in [v1.1.48](https://github.com/swisspush/gateleen/releases/tag/v1.1.48)

PR #294, opened 2019-10-09
Issue291 kafka module

mcweba:
PR for issue [#291 introduce gateleen-kafka module](https://github.com/swisspush/gateleen/issues/291)

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/configuration/ConfigurationResourceManager.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/configuration/ConfigurationResourceManager.java
@@ -1,7 +1,9 @@
 package org.swisspush.gateleen.core.configuration;
 
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.eventbus.Message;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.http.HttpServerRequest;
@@ -19,6 +21,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 /**
  * Manage the modifications of configuration resources and notify observers
@@ -54,7 +57,7 @@ public class ConfigurationResourceManager implements LoggableResource {
 
             if (requestUri != null && type != null) {
                 if (ConfigurationResourceChangeType.CHANGE == type) {
-                    notifyObserversAboutResourceChange(requestUri);
+                    notifyObserverAboutResourceChange(requestUri, null);
                 } else if (ConfigurationResourceChangeType.REMOVE == type) {
                     notifyObserversAboutRemovedResource(requestUri);
                 } else {
@@ -64,6 +67,16 @@ public class ConfigurationResourceManager implements LoggableResource {
                 log.warn("Invalid configuration resource change message received. Don't notify anybody!");
             }
         });
+    }
+
+    /**
+     * Call this method to get the validated registered resource immediately. Otherwise you would have to wait
+     * for a change on the corresponding resource.
+     *
+     * @param resourceUri the uri of the registered resource
+     */
+    public Future<Optional<Buffer>> getRegisteredResource(String resourceUri){
+        return getValidatedRegisteredResource(resourceUri);
     }
 
     public void registerResource(String resourceUri, String resourceSchema) {
@@ -164,6 +177,30 @@ public class ConfigurationResourceManager implements LoggableResource {
         return observers;
     }
 
+    private Future<Optional<Buffer>> getValidatedRegisteredResource(String resourceUri){
+        Future<Optional<Buffer>> future = Future.future();
+        String resourceSchema = getRegisteredResources().get(resourceUri);
+        storage.get(resourceUri, buffer -> {
+            if (buffer != null) {
+                configurationResourceValidator.validateConfigurationResource(buffer, resourceSchema, event -> {
+                    if (event.succeeded()) {
+                        if (event.result().isSuccess()) {
+                            future.complete(Optional.of(buffer));
+                        } else {
+                            future.fail("Failure during validation of resource " + resourceUri + ". Message: " + event.result().getMessage());
+                        }
+                    } else {
+                        future.fail("Failure during validation of resource " + resourceUri + ". Message: " + event.cause());
+                    }
+                });
+            } else {
+                future.complete(Optional.empty());
+            }
+        });
+
+        return future;
+    }
+
     private void notifyObserversAboutRemovedResource(String requestUri) {
         log.info("About to notify observers that resource " + requestUri + " has been removed");
         List<ConfigurationResourceObserver> observersByResourceUri = getObserversByResourceUri(requestUri);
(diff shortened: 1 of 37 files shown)


gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaProducerRecordBuilder.java
```diff
@@ -0,0 +1,101 @@
...
+     * @param topic the kafka topic
+     * @param payload the payload to build the {@link KafkaProducerRecord}s from
+     * @return A list of {@link KafkaProducerRecord}s created from the provided payload
+     * @throws ValidationException when the payload is not valid (missing properties, wrong types, etc.)
+     */
+    static List<KafkaProducerRecord<String, String>> buildRecords(String topic, Buffer payload) throws ValidationException {
+        List<KafkaProducerRecord<String, String>> kafkaProducerRecords = new ArrayList<>();
+        JsonObject payloadObj;
+        try {
+            payloadObj = new JsonObject(payload.toString());
```
hiddenalpha: Is it by intent using `toString()` with implicit charset?

gateleen-kafka/src/main/resources/gateleen_kafka_topic_configuration_schema
```diff
@@ -0,0 +1,35 @@
...
+  },
+  "definitions": {
+    "Topic": {
+      "description": "A single kafka topic entry",
+      "type": "object",
+      "additionalProperties": {
+        "type": "string"
+      },
+      "properties": {
+        "bootstrap.servers": {
```
hiddenalpha: Do we really want _dot_ in property name? IMHO a candidate to provocate confusion.

mcweba: Yes we do, because these are Kafka related properties which must be called exactly like defined in the schema. See [Kafka doc](https://kafka.apache.org/documentation/#producerconfigs)

gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaConfigurationParser.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Parses the provided kafka topic configuration resource and returns a list of {@link KafkaConfiguration} objects.
+     *
+     * When the configuration resource contains invalid regex patterns, a warning will be logged and the corresponding
+     * {@link KafkaConfiguration} object will not be included in the returned list.
+     *
+     * @param configurationResourceBuffer the resource to parse
+     * @return a list of {@link KafkaConfiguration} objects
+     */
+    static List<KafkaConfiguration> parse(Buffer configurationResourceBuffer) {
```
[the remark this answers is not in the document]
mcweba: The whole class is _package private_, because it's not needed outside the kafka package.

gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaHandler.java
```diff
@@ -0,0 +1,157 @@
...
+            }
+            initialized = true;
+            future.complete();
+        });
+
+        return future;
+    }
+
+    public boolean handle(final HttpServerRequest request) {
+        final Logger requestLog = RequestLoggerFactory.getLogger(KafkaHandler.class, request);
```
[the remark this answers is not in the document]
mcweba: Good point, will move it.

gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaProducerRecordBuilder.java
mcweba: The question should be: Why using _String_ at all. We can use the _Buffer_ directly. Will change this.

gateleen-core/src/main/java/org/swisspush/gateleen/core/configuration/ConfigurationResourceManager.java
```diff
@@ -164,63 +177,57 @@ public boolean handleConfigurationResource(final HttpServerRequest request) {
...
-                        if (event.result().isSuccess()) {
-                            observer.resourceChanged(requestUri, buffer.toString());
-                        } else {
-                            log.warn("Failure during validation of resource " + requestUri + ". Message: " + event.result().getMessage());
-                        }
-                    } else {
-                        log.warn("Failure during validation of resource " + requestUri + ". Message: " + event.cause());
+        getValidatedRegisteredResource(requestUri).setHandler(event -> {
+            if(event.failed()){
+                log.warn(event.cause().getMessage());
```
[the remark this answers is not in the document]
mcweba: The failureMessage already includes the cause (where present). See `future.fail("Failure during validation of resource " + resourceUri + ". Message: " + event.result().getMessage());` and `future.fail("Failure during validation of resource " + resourceUri + ". Message: " + event.cause());`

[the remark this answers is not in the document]
hiddenalpha: Therefore we're using _String_ to communicate errors rather than _Exception_. Not my taste about how to communicate errors. But that's only my opinion ;)

gateleen-core/src/main/java/org/swisspush/gateleen/core/future/SequentialFutures.java
```diff
@@ -0,0 +1,22 @@
...
+public class SequentialFutures {
+
+    public static <A, B> B execute(Iterator<A> iterator, B identity, BiFunction<B, A, B> function) {
+        B result = identity;
+        while(iterator.hasNext()) {
+            A next = iterator.next();
+            result = function.apply(result, next);
+        }
+        return result;
+    }
```
lbovet: This is basically a Stream.reduce(). Also, this code is actually not specific to futures.
Can we not replace this with:
```
list.stream().reduce(Future.succeededFuture(), (f, item) -> f.compose( action(item) ))
```
In the stackoverflow article, they did the combine (compose) after, which is the setup for parallelizing reduce. If we do it directly, it should work.

mcweba: You are right, I was able to replace the _SequentialFutures_ class with a "simple" _Stream.reduce()_

gateleen-core/src/main/java/org/swisspush/gateleen/core/configuration/ConfigurationResourceManager.java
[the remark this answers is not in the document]
mcweba: In fact, this could have been implemented "better". On the other side, possible errors are logged, which is what we want here.

However, I think we should focus such code reviews on the more important / mission critical code (and functionality) instead of minor "code styling" issues.

gateleen-core/src/main/java/org/swisspush/gateleen/core/event/EventBusHandler.java
```diff
@@ -240,9 +240,9 @@ public void install(Router router) {
     }
 
     @Override
-    public void resourceChanged(String resourceUri, String resource) {
+    public void resourceChanged(String resourceUri, Buffer resource) {
         if(configResourceUri() != null && configResourceUri().equals(resourceUri)){
-            log.info("Got notified about configuration resource update for "+resourceUri+" with new data: " + resource);
+            log.info("Got notified about configuration resource update for " + resourceUri);
```
hiddenalpha: I suggest to use [Parameterized Logging](https://www.tutorialspoint.com/slf4j/slf4j_parameterized_logging.htm). (There are also other occurences)

PR #301, opened 2020-03-02
Replaced "setAll" headers method with "addAll" headers method to prev…

mcweba:
…ent existing headers from being removed


oli-h:
Note that we need to discuss this first. Can have heavy impact on exising Gatelen-servers

PR #304, opened 2020-03-09
Discontinued to use LocalDateTime because of issues upon timezone change

dominik-cnx:
The route hooks currently expire we we enter daylight savings time (and probably do not expire on time when we leave daylight savings time). With this PR we use DateTime (which is timezone aware) to avoid such issues. 

I also removed some not used code.

### DIFF
--- a/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
+++ b/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
@@ -6,7 +6,6 @@ import io.vertx.core.http.HttpServerRequest;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Instant;
-import org.joda.time.LocalDateTime;
 import org.joda.time.format.DateTimeFormatter;
 import org.joda.time.format.ISODateTimeFormat;
 import org.slf4j.Logger;
@@ -32,7 +31,6 @@ public final class ExpiryCheckHandler {
 
     private static DateTimeFormatter dfISO8601 = ISODateTimeFormat.dateTime().withZone(DateTimeZone.forID("Europe/Zurich"));
     private static DateTimeFormatter dfISO8601Parser = ISODateTimeFormat.dateTimeParser().withZone(DateTimeZone.forID("Europe/Zurich"));
-    private static DateTimeFormatter isoDateTimeParser = ISODateTimeFormat.dateTimeParser();
 
     private ExpiryCheckHandler() {
         // Prevent instantiation
@@ -40,7 +38,7 @@ public final class ExpiryCheckHandler {
 
     /**
      * Checks if a "X-Server-Timestamp" header is set or not.
-     * If no valid value is found, the actual timestamp is set.
+     * If no valid value is found, the current timestamp is set.
      * 
      * @param request request
      */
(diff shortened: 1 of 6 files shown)


oli-h:
Note that the title is misleadung: We don't talk about "timezone change". We talk about "clock-jumps due to daylight savings" which happens twice a year

hiddenalpha:
Why did the [travis build](https://travis-ci.org/swisspush/gateleen/builds/659963389?utm_source=github_status&utm_medium=notification) fail?

dominik-cnx:
Daylight saving time results in timezone changes, for me it is the same.

dominik-cnx:
> Why did the [travis build](https://travis-ci.org/swisspush/gateleen/builds/659963389?utm_source=github_status&utm_medium=notification) fail?

Not sure, but it passed in the meantime.

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
```diff
@@ -78,14 +66,8 @@ public static void updateServerTimestampHeader(MultiMap headers) {
...
-            String updatedTimestamp = localizeTimestamp(serverTimestamp);
-
-            if (!updatedTimestamp.equals(serverTimestamp)) {
-                log.debug("Updating" + SERVER_TIMESTAMP_HEADER + " value from " + serverTimestamp + "  to " + updatedTimestamp);
-
-                headers.remove(SERVER_TIMESTAMP_HEADER);
-                headers.set(SERVER_TIMESTAMP_HEADER, updatedTimestamp);
-            }
+            headers.remove(SERVER_TIMESTAMP_HEADER);
+            headers.set(SERVER_TIMESTAMP_HEADER, printDateTime(DateTime.now()));
```
hiddenalpha: Is it by intent to remove that condition?

dominik-cnx: I was actually uncertain here. Hope that @ljucam can say something here.

Swisspusher:
What happens to already existing hooks with no timezone information? We have this exiprationTime in our hooks for example:
```
"expirationTime": "2020-03-10T11:24:00.861"
```

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
Swisspusher: Before approving we should be clear about the removal of this condition.

mcweba:
> What happens to already existing hooks with no timezone information? We have this exiprationTime in our hooks for example:
> 
> ```
> "expirationTime": "2020-03-10T11:24:00.861"
> ```

Just discussed with @dominik-cnx about this. The timezone (Europe/Zurich) will be added by the DateTimeFormatter, which is OK. 

On the other side we should maybe change the hard coded timezone to be configurable since Gateleen is used world wide ;-)

dominik-cnx:
Let's wait for @ljucam to approve the change as well.

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/expiry/ExpiryCheckHandler.java
ljucam: Honestly, I’m not really sure. Before this change, we updated the timestamp only then, when it differed to the current localized timestamp. Now we directly set it to the current timestamp. So, we actively always change the server timestamp, even if it would be right. Before that, we just converted it. Therefore, I’m not sure, what could change. 


```
private static String localizeTimestamp(String timestamp) {
        String localizedTimestamp = timestamp;
        if (localizedTimestamp != null && localizedTimestamp.toUpperCase().endsWith("Z")) {
            try {
                DateTime dt = isoDateTimeParser.parseDateTime(localizedTimestamp);
                localizedTimestamp = dfISO8601.print(dt);
            } catch (IllegalArgumentException e) {
                log.warn("Could not parse: " + localizedTimestamp);
            }
        }

        return localizedTimestamp;
    }
```

dominik-cnx: If we cannot explain what we need the equals() check for then I suggest that we remove it and do exactly what it says in the method description:

"If no valid value is found, the actual timestamp is set."

mcweba: I am OK with this

Issue #314, opened 2020-04-21
Add ReturnHttpErrorHandler


mcweba:
In the ISA project a new handler called _ReturnHttpErrorHandler_ was implemented to provide a better way for responding requests with a given status code.

Currently gateleen allows a _translateStatus_ property to be set on a routing rule. Since this only sets the status code and not the status message, you can get responses like:
> 503 OK

The implementation of the _ReturnHttpErrorHandler_ to adopt looks like this:

```java
public class ReturnHttpErrorHandler {

    private final Logger LOG = LoggerFactory.getLogger(ReturnHttpErrorHandler.class);

    private final String path;

    public ReturnHttpErrorHandler(String rootPath) {
        this.path = rootPath;
        LOG.info("listening to {}", rootPath);
    }

    public boolean handle(HttpServerRequest request) {
        if (request.uri().startsWith(path) == false) {
            return false;
        }
        HttpResponseStatus rs;
        String info = "";
        try {
            String code = request.uri().substring(path.length());
            while (code.startsWith("/")) {
                code = code.substring(1);
            }
            int codeAsInt = Integer.parseInt(code);
            rs = HttpResponseStatus.valueOf(codeAsInt);
        } catch (Exception ex) {
            LOG.warn("can't parse wanted respose code from {}", request.uri(), ex);
            rs = HttpResponseStatus.BAD_REQUEST;
            info = ": missing, wrong or non-numeric status-code in request URL";
        }
        request.response().setStatusCode(rs.code()).setStatusMessage(rs.reasonPhrase()).end(rs.toString() + info);
        return true;
    }

}
```
This handler allows routing rules which can look like this having configured a roothPath of _/gateleen/server/return-http-error_:
```
"/gateleen/server/unreachable/service": {
    "path": "/gateleen/server/return-http-error/503",
    "methods": [
      "GET"
    ]
}
```

mcweba:
Released in [v1.1.54](https://github.com/swisspush/gateleen/releases/tag/v1.1.54)

PR #317, opened 2020-04-23
PR for issue #314

mcweba:


gateleen-playground/src/main/resources/playground/server/admin/v1/routing/rules
```diff
@@ -1,4 +1,8 @@
 {
+  "/playground/service/unavailable/(.*)": {
```
oli-h, line 2: You don't need a capture group. I.e. it's enough to write `/playground/service/unavailable/.*`

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/CustomHttpResponseHandler.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public boolean handle(HttpServerRequest request) {
+        if (!request.uri().startsWith(path)) {
+            return false;
+        }
+        HttpResponseStatus rs;
+        String info = "";
+        try {
+            String code = request.uri().substring(path.length());
```
oli-h, line 30: @hiddenalpha proposed to isolate the "numeric" part of the URI-trailer, i.e. to also accept
`/gateleen/server/return-with-status-code/404/some/other/path/segments/which/are/ignored`

This gives us the option to work woth a Route-Hook where _destination is this Handler_.
Note that Route-Hooks always add the Path-Segments 'right of the hooked path' to the destination-URL

gateleen-playground/src/main/resources/playground/server/admin/v1/routing/rules
mcweba: Correct. Since all other rules in this resource have a capture group, I leave this as it is to keep consistency

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/CustomHttpResponseHandler.java
```diff
@@ -35,8 +34,8 @@ public boolean handle(HttpServerRequest request) {
         try {
             String code = request.uri().substring(path.length());
 
-            List<String> pathSegments = Lists.newArrayList(Splitter.on(SLASH).omitEmptyStrings().limit(2).split(code));
-            code = pathSegments.isEmpty() ? EMPTY : pathSegments.get(0);
+            final Iterator<String> pathSegments = Splitter.on(SLASH).omitEmptyStrings().limit(2).split(code).iterator();
```
oli-h, line 37: Uh - I only see now that this is Splitter from Guava - aaargh. Massive fat library for stuff which is (mostly) part of Java 8 :-/

mcweba: Already removed. Have another look

Issue #319, opened 2020-04-29
Add translateStatus to route hooks


mcweba:
_Regular_ routing rules can be configured to reply with custom http response status. This feature is not available for route hooks.

We have the problem that queue items which are forwarded by a route hook block the queue when failed to process. In the case of a 404 Not Found, we would like to discard the queue item and continue with the next.

hiddenalpha:
As a quick shot, I see no advantage how we (ISA) use gateleen in our project.

Usually we handle this case as:
1. A service MUST NOT order stuff he cannot process. (Do hook precisely. Eg using _filter_)
2. This is a consumer-side error and needs fix there.

But we also have queues to 3rd-party services where _fixing it there_ is not an easy option. In those cases we usually:
1. Use _X-Queue-Expire-After_
2. Additionally, we're monitoring queue sizes and trigger alarms in case of abnormally large queues.
3. This way we've a _fully-automated-bugs-detector_ and this even for free ;)

mcweba:
We also have queues to 3rd-party services and the provided options are not sufficient. As a workaround we also use the _X-Queue-Expire-After_ header.

However, there's a problem with this approach. Let's assume we have a first queue item _Q1_ with an _X-Queue-Expire-After_ configuration. This queue item is rejected by the 3rd-party service with a _404 Not Found_ response. Then we have a second queue item _Q2_ with a queue expiry configuration that's 1 millisecond after the queue expiry of _Q1_. So when _Q1_ finally is removed from the queue, _Q2_ will most likely already be expired when processed.
In addition to that, the queue item _Q2_ has to wait unnecessarily until _Q1_ is expired when already with the first attempt of processing _Q1_ is clear that this queue item should be removed.

mcweba:
@lbovet / @oli-h what's your opinion?

lbovet:
Conceptually, if translateStatus makes sense in static routing, it also makes sense in hook routing.
We could have only hook routing, after all...
Then the question is "does translate status makes sense at all?". The answer is yes, because we don't have full control on the status codes of downstream backends. Gateleen integrates external stuff and many of its features are there to adapt to this stuff.

mcweba:
Released in [v1.1.56](https://github.com/swisspush/gateleen/releases/tag/v1.1.56)

PR #326, opened 2020-06-24
#319 Add translateStatus to route hooks

mcweba:


gateleen-hook/src/main/resources/gateleen_hooking_schema_hook
```diff
@@ -37,6 +37,13 @@
 		"destination": {
 			"type": "string"
 		},
+        "translateStatus": {
+            "description": "Mapping to transform backend HTTP status.",
+            "type": "object",
+            "additionalProperties": {
+                "type": "number"
```
hiddenalpha, line 44: Why floating point number?

lbovet: For uncertain response codes like 301.5 (Redirect semi-permanently).

hiddenalpha: I've never seen that so far. Is there any spec describing this? (So I can learn about them) Only thing I could find was [RFC2616](https://www.ietf.org/rfc/rfc2616.txt). But I cannot see any hint for codes like 301.5. What am I missing?

> 6.1.1 Status Code and Reason Phrase
> 
> The Status-Code element **is a 3-digit integer** result code ...

lbovet: Oops, I forgot the smiley, it was irony :)

mcweba: Good one 🤣. Or like 201.5 Almost Accepted

PR #330, opened 2020-07-31
Make Forwarder let devs know when someting gets wrong

hiddenalpha:
We've a hard-to-discover bug on some of our machines where we have
gateleen in place. While trying to find the real cause we ended up at
this code, which seems to hide Throwables of any kind. This patch
(hopefully) will help to find the real cause behind our randomly occurring
problem.

Further, I cannot imagine any senseful reason why we should (nearly
silently, no stack, etc) ignore Throwables like:

NullPointerException, OutOfMemoryError, ThreadDeath, StackOverflowError,
AssertionError, IOError, ThreadDeath, ... (and more)

### DIFF
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -315,6 +315,7 @@ public class Forwarder implements Handler<RoutingContext> {
                 error("Timeout", req, targetUri);
                 respondError(req, StatusCode.TIMEOUT);
             } else {
+                LOG.warn("Failed to '{} {}'", req.method(), targetUri, exception);
                 error(exception.getMessage(), req, targetUri);
                 if (req.response().ended() || req.response().headWritten()) {
                     error("Response already written. Not sure about the state. Closing server connection for stability reason", req, targetUri);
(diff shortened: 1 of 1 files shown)


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -315,6 +315,7 @@ private void installExceptionHandler(final HttpServerRequest req, final String t
                 error("Timeout", req, targetUri);
                 respondError(req, StatusCode.TIMEOUT);
             } else {
+                LOG.warn("Failed to '{} {}'", req.method(), targetUri, exception);
```
ljucam, line 318: Missing words: Failed to read ... (?) ...

mcweba, line 318: `exception` parameter is newer used when only having two `{}` placeholders. Or do I miss something?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -385,6 +386,7 @@ private HttpClientRequest prepareRequest(final HttpServerRequest req, final Stri
             };
 
             cRes.exceptionHandler(exception -> {
+                LOG.warn("Failed to read upstream response for '{} {}'", req.method(), targetUri, exception);
```
mcweba, line 389: `exception` parameter is newer used when only having two `{}` placeholders. Or do I miss something?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
hiddenalpha: [how-to-log-exception-and-message-with-placeholders-with-slf4j](https://stackoverflow.com/q/5951209/4415884)

mcweba: Nice. They did a good job hiding this in the documentation ;-)

mcweba (approved):

PR #332, opened 2020-08-24
property mapper for gateleen-kafka

geekrumper:
Optional property mapper for the KafkaHandler

### DIFF
--- a/gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaHandler.java
+++ b/gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaHandler.java
@@ -92,26 +92,12 @@ public class KafkaHandler extends ConfigurationResourceConsumer {
 
     private Future<Void> initializeKafkaConfiguration(Buffer configuration) {
         Future<Void> future = Future.future();
-        String replacedConfig;
-        final List<KafkaConfiguration> kafkaConfigurations = new ArrayList<>();
-
-        try {
-            replacedConfig = StringUtils.replaceWildcardConfigs(configuration.toString(UTF_8), properties);
-            kafkaConfigurations.addAll(KafkaConfigurationParser.parse(Buffer.buffer(replacedConfig)));
-        } catch (Exception e) {
-            log.warn("Could not replace wildcards with environment properties for kafka configurations due to following reason: {}",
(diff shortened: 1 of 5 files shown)


mcweba (changes requested):

gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaHandler.java
```diff
@@ -72,7 +91,10 @@ public boolean isInitialized() {
 
     private Future<Void> initializeKafkaConfiguration(Buffer configuration) {
         Future<Void> future = Future.future();
-        final List<KafkaConfiguration> kafkaConfigurations = KafkaConfigurationParser.parse(configuration);
+        String replacedConfig;
+        replacedConfig = StringUtils.replaceWildcardConfigs(configuration.toString(UTF_8), properties);
```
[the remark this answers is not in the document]
geekrumper: Exception handling added. We still catch it, warn about the reason and set initialized to false without stopping the flow (e.g. rethrowing exception).

gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaHandler.java
```diff
@@ -92,14 +93,25 @@ public boolean isInitialized() {
...
+        } catch (Exception e) {
+            log.warn("Could not replace wildcards with environment properties for kafka configurations due to following reason: {}",
+                    e.getMessage());
+        }
+
         repository.closeAll().setHandler(event -> {
-            for (KafkaConfiguration kafkaConfiguration : kafkaConfigurations) {
-                repository.addKafkaProducer(kafkaConfiguration);
+            if (kafkaConfigurations.isEmpty()) {
+                initialized = false;
```
mcweba, line 108: empty configurations are not leading to an un-inititialized handler. I just noticed, that the wildcard replacing should be made in the `KafkaConfigurationParser` and not in the `KafkaHandler` -> separation of concerns (SoC)

geekrumper: implementation moved to `KafkaConfigurationParser`

mcweba (approved):

Issue #336, opened 2020-09-22
Restrict user specific resource access


mcweba:
### Problem
The current implementation of the gateleen-security module does allow to restrict access to specific URLs like ```/playground/server/users/v1/<USER_ID>/.*```

However, there's no possiblity to restrict user A from calling user B's profile with ```GET /playground/server/users/v1/B/profile``` when user A has the permission (by ACL) to access user profiles.

### Solution
Allow the ACLs to have paths with wildcard segments. These segments hold wildcard values for request headers.

Example:
```json
{
	"profile": {
		"methods": [
			"GET",
			"PUT"
		],
		"path": "/playground/server/users/v1/{x-user}/profile(\\?.*)?"
	}
}
```
When a request is processed, the ```{x-user}``` segment will be replaced with the value from the ```{x-user}``` http request header.

mcweba:
Released in [v1.1.58](https://github.com/swisspush/gateleen/releases/tag/v1.1.58)

PR #338, opened 2020-11-09
pull request for #336 restrict access to user specific resources

mcweba:


gateleen-security/README_security.md
```diff
@@ -63,4 +63,45 @@ Any request with a user group containing 'acl-domain-' will match the defined ro
...
\ No newline at end of file
+Note: The rolemapper object is validated against the gateleen_security_schema_rolemapper json schema
+
+#### Using request headers
+In ACLs you can use wildcard definitions in the urls. These wildcards can contain request header names and will be replaced with the actual
+values from a request.
+
+The wildcards have to be in the format 
+```
+<header name>
```
lbovet, line 74: What was the reason not to use { } which looks more "variablish"?

mcweba: I tried to use `{}` first but got massive problems with the Java Pattern class because normally in a regex pattern the `{}` is reserved for quantifiers like `{5}` or `{3,5}`. I could fix this by adding the `Pattern.LITERAL` option during pattern parsing. However, urls having such quantifiers did not work anymore. Example: `/resources/userspecific/{x-user}/items/([0-9]{6})/foo`. Switching to `<>` solved these problems

lbovet: Ah yes, forgot the quantifier.
Makes sense. < > is something we should not see in a normal URL indeed.

lbovet (approved):

stettlermari (approved):

gateleen-security/README_security.md
hiddenalpha: Be aware of [Named Capture Groups](https://riptutorial.com/regex/example/2479/named-capture-groups). Nevertheless I know no place where we would use them in URLs ;)

Issue #350, opened 2021-01-21
Make default behaviour of RoleAuthorizer configurable


mcweba:
The current implementation of the _RoleAuthorizer_ class grants access to requests without any roles. From a security perspective, this behaviour is a bit strange and can be summarized by the following statement.

> When you have roles, you must have the right roles to get access to a specific resource. When you don't have any roles at all, you are free to access whatever you want.

To keep backward compatibility, I would suggest provide a way to configure the default behaviour.

lbovet:
Yes, that is something like a "dev mode" flag.

mcweba:
Released in [v1.1.63](https://github.com/swisspush/gateleen/releases/tag/v1.1.63)

PR #351, opened 2021-01-21
#350 configure requests without roles to be granted or rejected

mcweba:


hiddenalpha (approved):
I don't know this module well. But "feels" good 👍

PR #368, opened 2021-03-19
Changed some logs from type info to type debug

mcweba:


stettlermari (approved):

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/reducedpropagation/ReducedPropagationManager.java
```diff
@@ -173,7 +173,7 @@ private void processExpiredQueues(String lockToken) {
                         if (!OK.equals(event1.result().body().getString(STATUS))) {
                             log.error("Failed to process expired queue '"+expiredQueue+"'. Message: " + event1.result().body().getString(MESSAGE));
                         } else {
-                            log.info(event1.result().body().getString(MESSAGE));
+                            log.debug(event1.result().body().getString(MESSAGE));
```
hiddenalpha, line 176: (*minor*) I think here it would make sense to use a guard:
`if(log.isDebugEnabled()){ ... }`
this would prevent from calling `getString` everytime. Which *might* be an expensive operation (I did not test it)

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/reducedpropagation/ReducedPropagationManager.java
```diff
@@ -105,7 +105,7 @@ public void startExpiredQueueProcessing(long intervalMs) {
 
         long expireTS = System.currentTimeMillis() + propagationIntervalMs;
 
-        log.info("Going to perform a lockedEnqueue for (original) queue '" + queue + "' and eventually starting a new timer");
+        log.debug("Going to perform a lockedEnqueue for (original) queue '" + queue + "' and eventually starting a new timer");
```
hiddenalpha, line 108: (*minor*) Should we replace all those string concatenations with placeholders? (multiple occurrences)

See also [SO question logger-slf4j-advantages-of-formatting-with-instead-of-string-concatenation](https://stackoverflow.com/questions/10555409/logger-slf4j-advantages-of-formatting-with-instead-of-string-concatenation)

hiddenalpha (approved):

hiddenalpha (approved):
Looks like a lot of work 👍 Honestly I did not review every single line ;)

Issue #373, opened 2021-06-08
Reduce DeltaHandler logs for not found collections


mcweba:
When making delta requests to a collection which does not exist, the following error is logged correctly:

```
2021-06-08 08:51:46,612 vnecri2 int  crush ERROR DeltaHandler - YL8Tgt7UPkSYyLF72OnNegAALQU Failed to handle get for collection
```
However, also the stacktrace of the corresponding _ResourceCollectionException_ is logged.

Example:
```
org.swisspush.gateleen.core.util.ResourceCollectionException: Not Found
	at org.swisspush.gateleen.core.util.ExpansionDeltaUtil.checkResponse(ExpansionDeltaUtil.java:254)
	at org.swisspush.gateleen.core.util.ExpansionDeltaUtil.verifyCollectionResponse(ExpansionDeltaUtil.java:219)
	at org.swisspush.gateleen.delta.DeltaHandler.lambda$null$10(DeltaHandler.java:252)
	at org.swisspush.gateleen.core.http.LocalHttpServerResponse$1.lambda$bodyHandler$0(LocalHttpServerResponse.java:69)
	at org.swisspush.gateleen.core.http.BufferBridge.doEnd(BufferBridge.java:75)
	at org.swisspush.gateleen.core.http.LocalHttpServerResponse.end(LocalHttpServerResponse.java:269)
	at org.swisspush.gateleen.routing.StorageForwarder.lambda$null$1(StorageForwarder.java:141)
	at io.vertx.core.eventbus.impl.EventBusImpl.lambda$convertHandler$3(EventBusImpl.java:348)
	at io.vertx.core.eventbus.impl.HandlerRegistration.deliver(HandlerRegistration.java:261)
	at io.vertx.core.eventbus.impl.HandlerRegistration.handle(HandlerRegistration.java:239)
	at io.vertx.core.eventbus.impl.EventBusImpl$InboundDeliveryContext.next(EventBusImpl.java:565)
	at io.vertx.core.eventbus.impl.EventBusImpl.lambda$deliverToHandler$5(EventBusImpl.java:524)
	at io.vertx.core.impl.ContextImpl.executeTask(ContextImpl.java:320)
	at io.vertx.core.impl.EventLoopContext.lambda$executeAsync$0(EventLoopContext.java:38)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
[...]

PR #374, opened 2021-06-09
#373 changed log type from error to info for not-found collections. A…

mcweba:
Changed log type from error to info for not-found collections. Also do not log the stacktrace in this case


hiddenalpha (commented):
👍

PR #376, opened 2021-06-14
Added a new logger to keep track on HttpClient.close()

dominik-cnx:
We occasionally see 

java.io.IOException: Stream closed

in our project and would like to track down if this is related to cleanup done within Gateleen. With this PR we introduce a dedicated logger to log cleanup related activity.


mcweba:
I don't see the benefit of adding an addtional logger. These logs are already logged (with the same log level). You can identify those logs by looking for `Cleaning up...` log statements

dominik-cnx:
I know but we are hesitant to make the existing loggers more verbose because it might flood our logfiles.

Issue #378, opened 2021-06-16
Killswitch for delta requests using limit and offset request parameters


mcweba:
In general, requests using the _limit_ and _offset_ request parameters are a bit risky. Because the underlying collection can change between two requests, the result of a _limit_ and _offset_ request can change.

When combining such _limit_ and _offset_ requests with the _delta_ feature, it gets even worse, because the _delta_ values are processed on the limited result from the _limit_ and _offset_ response. This can lead to resources which should be listed according the _delta_ value, but are not returned because off the _limit_ and _offset_ result.

To stop clients from using the _limit_ and _offset_ parameters in _delta_ requests, we want to add a killswitch (constructor property) to reject those requests.

The default behaviour should still allow such requests to not break the backward compatibility.

mcweba:
Released in [v1.1.72](https://github.com/swisspush/gateleen/releases/tag/v1.1.72)

PR #379, opened 2021-06-16
#378 Killswitch for delta requests using limit and offset request par…

mcweba:
…ameters

### DIFF
--- a/gateleen-delta/src/main/java/org/swisspush/gateleen/delta/DeltaHandler.java
+++ b/gateleen-delta/src/main/java/org/swisspush/gateleen/delta/DeltaHandler.java
@@ -27,6 +27,8 @@ import java.util.Set;
 public class DeltaHandler {
 
     private static final String DELTA_PARAM = "delta";
+    private static final String LIMIT_PARAM = "limit";
+    private static final String OFFSET_PARAM = "offset";
     private static final String DELTA_HEADER = "x-delta";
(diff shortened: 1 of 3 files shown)


gateleen-delta/src/main/java/org/swisspush/gateleen/delta/DeltaHandler.java
```diff
@@ -200,6 +213,14 @@ private Long extractNumberDeltaParameter(String deltaStringId, HttpServerRequest
         }
     }
 
+    private void respondLimitOffsetParameterForbidden(HttpServerRequest request, Logger log){
+        String errorMsg = "limit/offset parameter not allowed for delta requests";
+        request.response().setStatusCode(StatusCode.BAD_REQUEST.getStatusCode());
+        request.response().setStatusMessage(StatusCode.BAD_REQUEST.getStatusMessage());
+        request.response().end(errorMsg);
+        log.warn(errorMsg);
```
hiddenalpha, line 221: Possibly helpful to also log some trivial details to ease bug-hunting. Eg like:
```java
log.warn( "{}: {} {}", errorMsg, request.method(), request.uri());
```

mcweba: The used RequestLogger provides the unique request ID, which helps a lot in bug hunting.

PR #389, opened 2021-07-07
access control allow method PATCH for CORS

gedestroy:
In our project we also use the PATCH method for cross-domain calls.


dominik-cnx:
@mcweba kindly review, thx.

Issue #382, opened 2021-06-29
Invalid JSON reported in Routing-Rule editor for property replacement variables


Kusig:
If there is used  property replacement variable for a numeric JSON attribute, the REST Editor does mark this as invalid JSON. 

Example:
```
(/myserver/services)?/mylink(/$)?(.*)": {
    "description": "blabla",
    "url": "${my.link}$3",
    "proxyOptions": {
      "type": "HTTP",
      "host": "${outproxy.link}",
      "port": ${outproxy.port}
    }
  }
```

The error is basically right because this is indeed invalid JSON at this time because the property-variable will be replaced only at runtime.

Reference issue: https://jira.post.ch/browse/SDCISA-1202

Kusig:
In oder do define such objects always as valid JSON, I propose to define the numeric attribute with a slightly different syntax:

"port": "$${outproxy.link}"

This way we could easily change Gateleen StringUtils  to replace this special tag without double quotes resulting in:

"port": 1234

This would work the same way just out of the box for boolean attributes as well .  It is 100% backwards compatible without any major risks.

@lbovet  WDYT?

lbovet:
An alternative would be to accept also strings for the port. It is cheaper, we don't add a feature to implement.

Kusig:
In this case we would have to create our own Proxy Class in between because the current implementation directly uses the vert.x class which we can not change.  As well, it would only work for this dedicated case and not in general.  As the feature looks to be very simple to be implemented, I'll go on with that one and will create a PR about.

lbovet:
Or simply transform the value here: https://github.com/swisspush/gateleen/blob/develop/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RuleFactory.java#L142
We should be able to keep a decoupling between our configuration API and the internal classes we use from vert.x

Kusig:
That's where I was already but I am not happy with it and discardet it.

This would only resolve exactly and only this case and only for this object and ruleset. It would have to be implemented for any other case with the same problem (eg. already just a boolean would cause the same problem).

From my point of view we need a generic approach where we could signal the system to transform the attribute to the right JSON type before the JSON transformation to the corresponding object is executed.

lbovet:
If we want the cleanest solution, we should address the problem where it is, i.e. in the REST editor to make it aware of variables, (and also able to check the that the variable language is correct). I think there are possibility to enable multiple syntaxes with the ace editor.

mcweba:
[...]

PR #390, opened 2021-07-08
#382 Invalid JSON reported in routing-rule with improvements

Kusig:
Improved the replacement in order to  fix the Java backslash escaping hell.
Added tests to validate normal and numeric json property with a real json file


gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
```diff
@@ -233,18 +234,29 @@ private static String replacePlainStringWildcardConfigs(String contentWithWildca
...
+            return contentWithWildcards;
+        }
         Engine engine = new Engine();
-        engine.setExprStartToken("\"$${");
-        engine.setExprEndToken("}\"");
-        return wildcardReplacementEngineIgnition(engine, jsonWithWildcards, properties);
+        engine.setExprStartToken(JSON_NUMERIC_START_TOKEN);
+        engine.setExprEndToken(JSON_NUMERIC_END_TOKEN);
+        String contentWithoutWildcards = wildcardReplacementEngineIgnition(engine, contentWithWildcards, properties);
+        // Note: due to the implicit string conversions of Java we just lost our escapes here and we must add
```
hiddenalpha, line 255: (*trivial, SideNote*) Is this really "Java" doing this? Isn't it that 3rd-party "Engine" we're using?

hiddenalpha (approved):
Cannot really say much as I don't know the background well. But looks good.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
Kusig: No, as far as I know, it is really the runtime doing this BS, even the compiler does something as well.
See here:  https://www.ibm.com/support/pages/when-use-double-backslash-escape-special-character-regular-expression

Overall a big mess.

PR #393, opened 2021-07-29
support to configure maxWaitQueueSize on routes

dominik-cnx:
Currently incoming requests get queued up once the connection pool is full. There is no size limit for this queue so far. With this PR we make it configurable. The default will be "unbounded", this means that everything stays the same if not actually configured.


gateleen-hook/README_hook.md
```diff
@@ -31,6 +31,7 @@ Adding hooks requires a validation against a schema to be positive. Check the sc
 | listable           | no  | This property specifies if a route is listed if a GET is performed on its parent or not. The default is false or set by the value passed during the initialization. | 
 | translateStatus    | no  | This property defines a mapping to translate http respond status values. | 
 | connectionPoolSize | no  | This property defines the max number of TCP connections which will be open in parallel to the destination system (Default 50) | 
+| maxWaitQueueSize   | no  | This property defines the max number of requests allowed in the wait queue. (Default -1, unbounded) |
```
Kusig, line 34: It might be good to differentiate the word "queue" here to the gateleen queues as it is not the same.  
Here we talk about the waiting requests on the httpClient dispatcher.

dominik-cnx: I agree, will improve this.

mcweba (approved):

PR #396, opened 2021-08-13
Feature/java11

linovifian:
- Upgraded to java 11
- lifted some dependency version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: java
-dist: xenial
+dist: trusty
 jdk:
   - oraclejdk11
 services:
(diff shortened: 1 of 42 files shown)


ljucam (approved):

Swisspusher (approved):

dominik-cnx (approved):

mcweba:
I was just one minute late, but could you also change the versioning? I think we should go from v1.1.XX to v1.2.XX since the Java 11 change maybe also needs infastructure changes

linovifian:
@mcweba addressed it in [this commit](https://github.com/swisspush/gateleen/commit/66559cda0ac9232fd5026b0587891951ae80d0ec)

Issue #394, opened 2021-07-30
Preserve host header and allow  route specific configuration


Kusig:
At the moment, Gateleen does force the http Host header and fills in the the target host:port no matter what is existing or configured.

In order to provide proper Host entries, this behaviour must be improved.
1) Don't overwrite Host header if it is existing already
2) Fill in target host:post if Host header is missing
2) Allow to configure host header with routing rules headers entry

See SDCISA-6298

hiddenalpha:
*WARN* the `Host` header has lots of behavior specified by the HTTP protocol. See eg [rfc2616_sec14.23](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23), [rfc2616_sec5.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.2), [rfc2616_sec19.6.1.1](https://www.w3.org/Protocols/rfc2616/rfc2616-sec19.html#sec19.6.1.1).

We **MUST** make sure we do not violate any of them.

Kusig:
Yep, this is mostly known but not really treated today by Gateleen anyhow. It simply just overwrites any Host header with the target host which sounds rather strange. 

However, I will not remove this but just use it as fallback if there is no incoming host (from Proxy or Browser) or if there is no specific host in the route configured.

The assumption is that most of the backends behind gateleen just don't care about this Host:xxx heaeder currently. That's why this never popped up so far.

mcweba:
Released in [v1.1.77](https://github.com/swisspush/gateleen/releases/tag/v1.1.77)

PR #397, opened 2021-08-31
Revert "#394 Preserve Host header value if it exists"

dominik-cnx:
This reverts commit 8a00cbad, this is due to side effects seen: 

We have now cases where the host header doesn't match with the host in the URL (it actually carries the value "localhost"), this leads to rejection with a 503 on the receiving end (in this case an open shift platform).


ljucam (approved):

dominik-cnx:
There are issues with the build, I'll have a look.

PR #398, opened 2021-09-14
#394 Improvement for Host header handling

Kusig:
This is a reduced less invasive solution to the reverted PR https://github.com/swisspush/gateleen/pull/395

The host header will be now always overwritten with the target address as it was before except if there is an explicit Host header given in the routing rules.

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -42,9 +42,8 @@ public class HttpHeaderUtil {
      * @return The found header value or null if none found
      */
     public @Nullable static <T extends MultiMap> String getHeaderValue(@Nonnull T headers, @Nonnull String headerKey) {
-        String matchKey = headerKey.toLowerCase();
         for (Entry<String, String> entry : headers.entries()) {
-            if (entry.getKey().equalsIgnoreCase(matchKey)) {
+            if (entry.getKey().equalsIgnoreCase(headerKey)) {
                 String value = entry.getValue();
                 if (value != null) {
                     return value.trim();
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -2,6 +2,8 @@ package org.swisspush.gateleen.core.util;
 
 import io.vertx.core.MultiMap;
 import java.util.Map.Entry;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
 
 public class HttpHeaderUtil {
@@ -39,10 +41,10 @@ public class HttpHeaderUtil {
      *                  the key searching is non case sensitive.
      * @return The found header value or null if none found
      */
-    public static <T extends MultiMap> String getHeaderValue(T headers, String headerKey) {
+    public @Nullable static <T extends MultiMap> String getHeaderValue(@Nonnull T headers, @Nonnull String headerKey) {
         String matchKey = headerKey.toLowerCase();
         for (Entry<String, String> entry : headers.entries()) {
-            if (entry.getKey().toLowerCase().equals(matchKey)) {
+            if (entry.getKey().equalsIgnoreCase(matchKey)) {
                 String value = entry.getValue();
                 if (value != null) {
                     return value.trim();
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -1,6 +1,7 @@
 package org.swisspush.gateleen.core.util;
 
 import io.vertx.core.MultiMap;
+import java.util.Map.Entry;
 
 
 public class HttpHeaderUtil {
@@ -28,6 +29,28 @@ public class HttpHeaderUtil {
         headers.remove(CONNECTION);
 
         return headers;
+    }
+
+    /**
+     * Helper method to find the first occurrence of a http header within the given List of headers.
+     *
+     * @param headers   The Map with the header key - value pairs to be evaluated
+     * @param headerKey The key for the header pair we are searching for in the given map. Note that
+     *                  the key searching is non case sensitive.
+     * @return The found header value or null if none found
+     */
+    public static <T extends MultiMap> String getHeaderValue(T headers, String headerKey) {
+        String matchKey = headerKey.toLowerCase();
+        for (Entry<String, String> entry : headers.entries()) {
+            if (entry.getKey().toLowerCase().equals(matchKey)) {
+                String value = entry.getValue();
+                if (value != null) {
+                    return value.trim();
+                }
+                return null;
+            }
+        }
+        return null;
     }
 
 }
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -212,25 +212,21 @@ public class Forwarder implements Handler<RoutingContext> {
             cReq.headers().set("Authorization", "Basic " + base64UsernamePassword);
         }
 
-
         MultiMap headers = cReq.headers();
-        String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        final String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
         final HeaderFunctions.EvalScope evalScope = rule.getHeaderFunction().apply(headers);
-        String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        final String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
         // see https://github.com/swisspush/gateleen/issues/394
-        if (hostHeaderAfter == null ||
-            (hostHeaderAfter != null && hostHeaderAfter.equals(hostHeaderBefore))) {
-            // there was no host header before or the host header was not updated by the rule given, therefore
-            // the host header will be forced overwritten always independent of the incoming value.
-            // This allows us to configure for certain routings to external url a dedicated Host header
-            // which will not be overwritten.
-            //
-            // https://jira.post.ch/browse/NEMO-1494
-            // the Host has to be set, if only added it will add a second value and not overwrite existing ones
-            String newHost = target.split("/")[0];
+        if (hostHeaderAfter == null || hostHeaderAfter.equals(hostHeaderBefore)) {
+            // there was no host header before or the host header was not updated by the rule given,
+            // therefore it will be forced overwritten always independent of the incoming value.
+            final String newHost = target.split("/")[0];
             headers.set(HOST_HEADER, newHost);
             log.debug("Host header replaced by default target value: {}", newHost);
         } else {
+            // the host header was changed by the configured routing and therefore
+            // it is not updated. This allows us to configure for certain routings to external
+            // url a dedicated Host header which will not be overwritten.
             log.debug("Host header replaced by rule value: {}", hostHeaderAfter);
         }
 
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -184,6 +184,38 @@ public class Forwarder implements Handler<RoutingContext> {
         }
     }
 
+    /**
+     * Execute the HeaderFunctions chain and apply the configured rules headers therefore
+     * @param log The logger to be used
+     * @param headers The headers which must be updated according the current forwarders rule
+     * @return null if everything is properly done or a error message if something went wrong.
+     */
+    String applyHeaderFunctions(final Logger log, MultiMap headers) {
+        final String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        final HeaderFunctions.EvalScope evalScope = rule.getHeaderFunction().apply(headers);
+        final String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        // see https://github.com/swisspush/gateleen/issues/394
+        if (hostHeaderAfter == null || hostHeaderAfter.equals(hostHeaderBefore)) {
+            // there was no host header before or the host header was not updated by the rule given,
+            // therefore it will be forced overwritten independent of the incoming value if necessary.
+            final String newHost = target.split("/")[0];
+            if (newHost != null && !newHost.isEmpty() && !newHost.equals(hostHeaderAfter)) {
+                headers.set(HOST_HEADER, newHost);
+                log.debug("Host header {} replaced by default target value: {}",
+                    hostHeaderBefore,
+                    newHost);
+            }
+        } else {
+            // the host header was changed by the configured routing and therefore
+            // it is not updated. This allows us to configure for certain routings to external
+            // url a dedicated Host header which will not be overwritten.
+            log.debug("Host header {} replaced by rule value: {}",
+                hostHeaderBefore,
+                hostHeaderAfter);
+        }
+        return evalScope.getErrorMessage();
+    }
+
     private void handleRequest(final HttpServerRequest req, final Buffer bodyData, final String targetUri, final Logger log, final Map<String, String> profileHeaderMap) {
         final LoggingHandler loggingHandler = new LoggingHandler(loggingResourceManager, req, vertx.eventBus());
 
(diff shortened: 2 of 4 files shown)


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -211,9 +212,27 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
             cReq.headers().set("Authorization", "Basic " + base64UsernamePassword);
         }
 
-
         MultiMap headers = cReq.headers();
+        String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
         final HeaderFunctions.EvalScope evalScope = rule.getHeaderFunction().apply(headers);
+        String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        // see https://github.com/swisspush/gateleen/issues/394
+        if (hostHeaderAfter == null ||
```
hiddenalpha, line 220: Somehow this block here doesn't make sense to me. Especially those points are unclear to me:

- Why do we already set the host header to our target value beforehand [a few lines above](https://github.com/swisspush/gateleen/blob/v1.2.2/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java#L209)? So isn't our use-case already implemented by this line?
- IntelliJ inspector says that `hostHeaderAfter != null` is _"always true when reached"_ So do we really need that check?
- AFAIK this makes the headerFunctions useless for "Host" header as it strictly overrides the setting which the headerFunctions did. So why do we ever configure headerFunctions just to revert them in a hardcoded way? Why not just remove such (bad) headerFunctions in the first place from our config?

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -30,4 +31,26 @@
...
 
+    /**
+     * Helper method to find the first occurrence of a http header within the given List of headers.
+     *
+     * @param headers   The Map with the header key - value pairs to be evaluated
+     * @param headerKey The key for the header pair we are searching for in the given map. Note that
+     *                  the key searching is non case sensitive.
+     * @return The found header value or null if none found
+     */
+    public static <T extends MultiMap> String getHeaderValue(T headers, String headerKey) {
```
dominik-cnx, line 42: We should annotate this with @Nullable ?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -211,9 +212,24 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
             cReq.headers().set("Authorization", "Basic " + base64UsernamePassword);
         }
 
-
         MultiMap headers = cReq.headers();
+        final String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
         final HeaderFunctions.EvalScope evalScope = rule.getHeaderFunction().apply(headers);
+        final String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        // see https://github.com/swisspush/gateleen/issues/394
+        if (hostHeaderAfter == null || hostHeaderAfter.equals(hostHeaderBefore)) {
```
dominik-cnx, line 220: Is it possible to write a unit (or integration) test for the different scenarios?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig: The check is already optimized in the newest update. 
The HostHeader will only be set if there is a rule containing such a header config.  That's why we check this before and after the apply() to evaluate if it changed during this really. If it did change, we don't want to it to be overwritten by the default.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -211,9 +212,27 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
...
+            // there was no host header before or the host header was not updated by the rule given, therefore
+            // the host header will be forced overwritten always independent of the incoming value.
+            // This allows us to configure for certain routings to external url a dedicated Host header
+            // which will not be overwritten.
+            //
+            // https://jira.post.ch/browse/NEMO-1494
+            // the Host has to be set, if only added it will add a second value and not overwrite existing ones
+            String newHost = target.split("/")[0];
+            headers.set(HOST_HEADER, newHost);
+            log.debug("Host header replaced by default target value: {}", newHost);
```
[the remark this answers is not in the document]
Kusig: Will do

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
Kusig: Yes, and with Java 11 we could even as well use @Nonnull as well now

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig: This will become very tricky as the handleRequest method is huge and must be split up to make a dedicated test just for this. Do you have any suggestions?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig: BTW, the harcoded part is the old default which is magically working.  So the intention with this reduced set is "don't touch a running system". This means we do only apply the Host header for routes where this is explicitely configured. In all other cases the very old hardcoded default is still doing it's job.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
hiddenalpha: Good point 👍 I think we definitively need tests which demonstrate the use-cases we want to cover with this feature. Because the code still lets tons of questions open for me.

BTW. [extract-method](https://www.martinfowler.com/bliki/FunctionLength.html) could help. Eg:
```java
EvalScope applyHeaderFunctions( HttpClientRequest req ){
    MultiMap headers = req.headers();
    // ...
    evalScope = rule.getHeaderFunction().apply( headers );
    // ...
    if( thereIsAReasonToDoStrangeThings() ){
        headers.set( HOST_HEADER, newHost );
    }
    // ...
    if( evalScopeHasErrors() ){
        return evalScope;
    }
    return null;
}
```

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig: Ah now I see what you meant. This was a failed merge, I will correct this.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig: That's exactly what I am working on actually. But it is not so simple ...

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
```diff
@@ -0,0 +1,132 @@
...
+    Forwarder forwarder = new Forwarder(vertx, httpClient, rule, storage, loggingResourceManager,
+        monitoringHandler, USER_PROFILE_PATH);
+    MultiMap reqHeaders = new VertxHttpHeaders();
+    reqHeaders.add(HOST_HEADER, DEFAULTHOST + ":" + DEFAULTPORT);
+    String errorMessage = forwarder.applyHeaderFunctions(logger, reqHeaders);
+    Assert.assertNull(errorMessage);
+    Assert.assertEquals(reqHeaders.get(HOST_HEADER),DEFAULTHOST + ":" + DEFAULTPORT);
+  }
+
+
```
hiddenalpha, line 131: **WARN**: I think we should add the following test to the mix:

```java
@Test
public void hostIsAcceptedIfBothOfInputAndHeaderFunctionsWishToUseTheSameHost() {
    Rule rule = extractRule("/ruleWithHostHeader");
    Forwarder forwarder = new Forwarder(vertx, httpClient, rule, storage, loggingResourceManager,
            monitoringHandler, USER_PROFILE_PATH);
    MultiMap reqHeaders = new VertxHttpHeaders();
    reqHeaders.add(HOST_HEADER, "newHost");
    String errorMessage = forwarder.applyHeaderFunctions(logger, reqHeaders);
    Assert.assertNull(errorMessage);
    Assert.assertEquals("newHost", reqHeaders.get(HOST_HEADER));
}
```

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
```diff
@@ -0,0 +1,132 @@
...
+  @Test
+  public void testHeaderFunctionsDefaultWithoutHostHeader() {
+    Rule rule = extractRule("/ruleWithoutHeader");
+    Forwarder forwarder = new Forwarder(vertx, httpClient, rule, storage, loggingResourceManager,
+        monitoringHandler, USER_PROFILE_PATH);
+    MultiMap reqHeaders = new VertxHttpHeaders();
+    reqHeaders.add(HOST_HEADER, DEFAULTHOST + ":" + DEFAULTPORT);
+    String errorMessage = forwarder.applyHeaderFunctions(logger, reqHeaders);
+    Assert.assertNull(errorMessage);
+    Assert.assertEquals(reqHeaders.get(HOST_HEADER),DEFAULTHOST + ":" + DEFAULTPORT);
```
hiddenalpha, line 128: > Assert.assertEquals(reqHeaders.get(HOST_HEADER),DEFAULTHOST + ":" + DEFAULTPORT);

We should swap 1st and 2nd argument. Because 1st arg is the "*expected value*" and 2nd arg is "*actual value*".

**HINT**: Same for all tests.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+
+}
```
hiddenalpha, line 132: *trivial*: I think would be nice to have a test which would test the "*return-an-error-string*" case. But I had no simple idea how to "*just-write*" such a test.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
```diff
@@ -0,0 +1,132 @@
...
+      rule.setHeaderFunction(headerFunction);
+    }
+    rule.setUrlPattern(extractedRule.getString("url"));
+    rule.setHost(DEFAULTHOST);
+    rule.setPort(DEFAULTPORT);
+    return rule;
+  }
+
+  @Before
+  public void setUp() {
```
dominik-cnx, line 74: Using the MockitoJUnitRunner possibly makes things easier

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
dominik-cnx: Minor, affects a few places, space missing
DER),DEFA

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
```diff
@@ -0,0 +1,132 @@
...
+    JsonArray headers = extractedRule.getJsonArray("headers");
+    if (headers != null) {
+      HeaderFunction headerFunction = HeaderFunctions.parseFromJson(headers);
+      rule.setHeaderFunction(headerFunction);
+    }
+    rule.setUrlPattern(extractedRule.getString("url"));
+    rule.setHost(DEFAULTHOST);
+    rule.setPort(DEFAULTPORT);
+    return rule;
+  }
```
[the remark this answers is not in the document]
Kusig: No idea, I'm going to ask my IDE ...

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
Kusig: I could not follow why this should make things easier.

dominik-cnx: You could use @Mock to create your mocks.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
Kusig: This is a good catch. But it is a wrong assumption. We do not care anymore the header coming in from somewhere. Therefore the case where the incoming header matches a configured header doesn't make sense as it must be set to the default in this case.   Or with other words, the incoming Host header for this method could only be the default, anything else must result in the default and configuring in the rule a host header which is the same as the default doesn't make sense really as it again results in the default.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
Kusig: Did not find a way to test this properly and to force failing the evaluation chain.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -30,4 +33,26 @@
...
+    /**
+     * Helper method to find the first occurrence of a http header within the given List of headers.
+     *
+     * @param headers   The Map with the header key - value pairs to be evaluated
+     * @param headerKey The key for the header pair we are searching for in the given map. Note that
+     *                  the key searching is non case sensitive.
+     * @return The found header value or null if none found
+     */
+    public @Nullable static <T extends MultiMap> String getHeaderValue(@Nonnull T headers, @Nonnull String headerKey) {
+        String matchKey = headerKey.toLowerCase();
```
hiddenalpha, line 45: Useless call to `.toLowerCase()` as we compare using `.equalsIgnoreCase()` anyway.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderTest.java
Kusig: As all other tests do run with the VertXUnitRunner I would like to keep it as it is in order to be in line with the other stuff which is already there..

hiddenalpha (approved):

ljucam (approved):

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -176,6 +184,38 @@ private String extractUserId(HttpServerRequest request, Logger log) {
...
+     */
+    String applyHeaderFunctions(final Logger log, MultiMap headers) {
+        final String hostHeaderBefore = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        final HeaderFunctions.EvalScope evalScope = rule.getHeaderFunction().apply(headers);
+        final String hostHeaderAfter = HttpHeaderUtil.getHeaderValue(headers, HOST_HEADER);
+        // see https://github.com/swisspush/gateleen/issues/394
+        if (hostHeaderAfter == null || hostHeaderAfter.equals(hostHeaderBefore)) {
+            // there was no host header before or the host header was not updated by the rule given,
+            // therefore it will be forced overwritten independent of the incoming value if necessary.
+            final String newHost = target.split("/")[0];
```
oli-h, line 201: Please note that `String#split()` is an expensive operation. Here you potentially extract the same hostname over and over again for every request.
Additionally I presume that the instance variable `target` only and always contains `host:port` - i.e. never something like an URL or URI. So you will _never_ find a forward `/` (though I'm not 100% sure). Therefore you probably can get rid of `String#split()`

hiddenalpha: Hi oli :) good catch. I think alternatively this operation could also be done in the constructor. Then we could hold the "newHost" in a field and re-use the same string instance each time. I think this could help too.

Kusig: Agree. Note that this was already before executed each time. It shouldn't be worse than before therefore.

oli-h: Yes, seems that "port" can be dynamic (i.e. potentially change for each request) - still no need to search for a forward-slash as there is none ;-)

But thinking on this "dynamic/wildcard port"-stuff: Is currrent implementation really robust here? Note that one Forwarder-instance is create for one routing-rule (or Hook) but then used for multiple request. Be careful with an argument that Vert.x is single-threaded: yes it is, still different methods on one Forwarder-instance are called at different times during a request/response-cycle.
```int port``` should probably be _not_ an instance variable

PR #403, opened 2021-11-15
Fix EventLoop-hog in ExpansionHandler

hiddenalpha:
We observed performance collapses for unrelated http clients while there
are expand requests running somewhere else. AFAIK this already is a
well-known, relatively old issue. I have no idea why It never got fixed
yet.

This introduces a SlicedLoop where the ExpansionHandler makes use
from. Based on my observations and experiments, the problem is that
expansion enqueues hundreds (if not thousands) of tasks to the
event-loop reactor. Further every such task tends to take a comparably
long time to complete.

The thresholds I preconfigured originate from my experiments where I
tried to tweak the misc values up and down until the performance
"looked ok".


gateleen-core/src/main/java/org/swisspush/gateleen/core/util/SlicedLoop.java
```diff
@@ -0,0 +1,208 @@
...
+import io.vertx.core.http.HttpClientRequest;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Iterator;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static java.lang.System.nanoTime;
+
+public class SlicedLoop<T> {
```
dominik-cnx, line 13: Propose to describe on class level what this is.

dominik-cnx (commented):
Is it possible to add a test the shows that it actually works / makes a difference?

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/SlicedLoop.java
```diff
@@ -0,0 +1,208 @@
...
+     */
+    public void pause() {
+        pauseRequest = true;
+    }
+
+    /**
+     * Resume (or start) the paused iteration. Think for {@link HttpClientRequest#resume()}.
+     */
+    public void resume() {
+        if (isRunning) throw new IllegalStateException("Already running");
```
ljucam, line 111: Please use block-syntax.

ljucam (approved):

mcweba (commented):
Do you have a reproducer (or unit test) where we can see how the new changes improve the performance?

mcweba:
> Do you have a reproducer (or unit test) where we can see how the new changes improve the performance?

I just saw that Dominik already wrote the same comment :-)

hiddenalpha:
> Is it possible to add a test the shows that it actually works / makes a difference?

Badly I didn't have a usable idea how to showcase this with a test as it requires lots of pre-conditions and setup to be met. Also it involves timing and this just leads to shaky tests as build-servers usually are not good at executing tests in similar timing conditions as prod or dev systems would.

While development I utilized a hand full of nodejs scripts performing requests against my gateleen where I manually did setup the pre-conditions needed in the storage for this to show up.

I'll take a look if I can bundle a reproducer and attach it here somehow.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/SlicedLoop.java
hiddenalpha: Good point. While trying to describe it, I found that the class by fact did multiple things at the same time. I now split off another class and now it was easier to describe what they're doing.

hiddenalpha:
Update regarding the reproducer:

I've still no good idea for a test. As a 1st step I'll explain how my reproducer looks like.

For this test I need THREE external components (see below). 'myClient' performs requests after each other via a route to 'nullsink' and measures how long those requests take. In normal circumstantces I observed values around 1-9 milliseconds.

Then 'myExpand' performs a hand full (eg 100 or so) '...?expand=...' requests to a subtree containing several hundred resources each 1 to 8 kB in size. As soon those requests start, 'myClient' reports a strong increase of response times. On a single expand request up to 500ms or under load even up to around 30 seconds.

This is pretty easy to setup as a simple reproducer. But right now I see no simple way to translate this setup into a automated test.

```
        +----------+
        | nullsink |
        +----------+
             A
             |  Route to some HTTP server.
             |
        +----------+
  .---->| gateleen |<----------.
  |     +----------+           |
  |                            |
  | Performs requests          | Performs a hand full of expensive expand
  | to 'nullsink' infinitely   | request. Eg to subtrees with a few
  | and measures time.         | hundred resources.
  |                            |
+----------+                +----------+
| myClient |                | myExpand |
+----------+                +----------+
```

mcweba:
> Update regarding the reproducer:
> 
> I've still no good idea for a test. As a 1st step I'll explain how my reproducer looks like.
> 
> For this test I need THREE external components (see below). 'myClient' performs requests after each other via a route to 'nullsink' and measures how long those requests take. In normal circumstantces I observed values around 1-9 milliseconds (localhost).
> 
> Then 'myExpand' performs a hand full (eg 100 or so) '...?expand=...' requests to a subtree containing several hundred resources each 1 to 8 kB in size. As soon those requests start, 'myClient' reports a strong increase of response times. On a single expand request up to 500ms or under load even up to around 30 seconds.
> 
> This is pretty easy to setup as a simple reproducer. But right now I see no simple way to translate this setup into a automated test.
> 
> ```
>         +----------+
>         | nullsink |
>         +----------+
>            A
>            | Route to some HTTP server.
>            |
>            |         +-------------------------+
>            |     .-->| Storage (redis) with a  |
>            |     |   | large subtree somewhere |
>            |     |   +-------------------------+
>            |     |
>         +----------+
>   .---->| gateleen |<----------.
>   |     +----------+           |
>   |                            |
>   | Performs requests          | Performs a hand full of expensive expand
>   | to 'nullsink' infinitely   | request. Eg to subtrees with a few
>   | and measures time.         | hundred resources.
>   |                            |
> +----------+                +----------+
> | myClient |                | myExpand |
> +----------+                +----------+
> ```

With the fix the response times of "myClient" do not increase (this much)? For me it is ok to not have a automated test.

lbovet:
The problem here is a backpressure issue. We flood the event queue much faster than it is processed. Other events cannot be interleaved and are queued after the whole bunch.
This proposed solution introduces a custom task scheduler with artificial delay which requires tuning for adjusting the priority between expand speed and latency of other requests. It is probably difficult to adjust in an optimal way.
Ideally, we should not have to tune this and let the system balance automatically. A simpler alternative would be to treat the loop items by batches and process the next batch only when all items of the previous have been processed. The batch size is proportional to the number of cores. This way, we let other events interleave and the implementation is simpler, since it is just about waiting on a multiple of futures.

mcweba:
About backpressure. Could this be targeted by using RxJava ([https://vertx.io/docs/vertx-rx/java2/](https://vertx.io/docs/vertx-rx/java2/))?

lbovet:
> About backpressure. Could this be targeted by using RxJava (https://vertx.io/docs/vertx-rx/java2/)?

Absolutely

hiddenalpha:
So how to continue here?

lbovet:
Would you like to try replacing the loop with a rx pipeline?
The flatMap operator has a parameter to choose the parallelism to limit the number of inflight sub-requests.
In fact, the rx pipeline implements the queuing behavior for us. It will be about 6 lines of code at the end.

hiddenalpha:
Recently found another time-slice and bundled a QuickNDirty reproducer together from my notes if there's still interest for it. Unluckily still takes some time to get it running (*uses nodejs*):
[reproducer-20220120-221633Z-gd61434c.tar.gz](https://github.com/swisspush/gateleen/files/7908882/reproducer-20220120-221633Z-gd61434c.tar.gz)
For setup instructions see bundled README and may refer to the [ASCII-drawing I wrote in comment_972722011](https://github.com/swisspush/gateleen/pull/403#issuecomment-972722011) to easier get a "*big picture*" how it is working.

PR #406, opened 2021-11-17
synchronize access to jmte Engine to avoid rare threading issue

dominik-cnx:
This is basically a workaround for something that should be addressed in https://github.com/HubSpot/jmte. However, there has been no activity in this project for the last 7 years or so. 

Also, there is some low level byte code compilation going on in jmte, not sure if this will continue to work with future Java versions. Actually I'm also uncertain if it does with current ones, I believe we do not use this feature in gateleen, so we don't necessarily see issues (and also do not necessarily care if there are issues).

We might want to look into alternatives to jmte to get rid of the dependency and this workaround.

Note: Also contains https://github.com/swisspush/gateleen/pull/405 to hopefully get a successful build.

### DIFF
--- a/gateleen-core/pom.xml
+++ b/gateleen-core/pom.xml
@@ -72,6 +72,7 @@
         <dependency>
             <groupId>org.testng</groupId>
             <artifactId>testng</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
     <build>
--- a/gateleen-core/pom.xml
+++ b/gateleen-core/pom.xml
@@ -69,6 +69,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
@@ -211,19 +211,21 @@ public final class StringUtils {
      * @return the String with replaced wildcard values. Returns the input String when input String or
      * properties are <code>null</code>
      */
-    private static synchronized String replacePlainStringWildcardConfigs(String contentWithWildcards,
+    private static String replacePlainStringWildcardConfigs(String contentWithWildcards,
         Map<String, Object> properties) {
         if (properties == null || contentWithWildcards == null) {
             return contentWithWildcards;
         }
         /*
-         * Note that this Engine is meant to be thread safe but it actually isn't because the TokenStream used by
-         * the Engine uses a MiniParser which in turn is not thread, see:
+         * Note that the Engine is meant to be thread safe but it actually isn't because the TokenStream used by
+         * the Engine uses a MiniParser which in turn is not thread safe, see:
          * https://github.com/HubSpot/jmte/blob/a1177fd89858ef55bb1d92acb4b1bb78bb7cac25/src/com/floreysoft/jmte/token/TokenStream.java#L39
          * Hence we synchronize access in this method.
          */
-        Engine engine = new Engine();
-        return wildcardReplacementEngineIgnition(engine, contentWithWildcards, properties);
+        synchronized (Engine.class) {
+            Engine engine = new Engine();
+            return wildcardReplacementEngineIgnition(engine, contentWithWildcards, properties);
+       }
     }
 
     /**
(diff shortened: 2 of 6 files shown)


gateleen-core/pom.xml
```diff
@@ -69,6 +69,10 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
```
ljucam, line 73: So, now we have junit and testng on board. Do we really want this?

ljucam (approved):

gateleen-core/pom.xml
dominik-cnx: Hmm, not so ideal I agree but

image

ljucam: Yeah, testng has a few advantages over junit.
Fine for me, as long as we are aware of this not-so-ideal situation.

pom.xml
```diff
@@ -343,6 +343,11 @@
                 <artifactId>junit</artifactId>
                 <version>4.13.1</version>
             </dependency>
+            <dependency>
+                <groupId>org.testng</groupId>
+                <artifactId>testng</artifactId>
+                <version>7.4.0</version>
```
mcweba, line 349: Shouldn't we have test dependencies in the `<scope>test</scope>`? This also applies to the junit dependency

mcweba:
What about updating _jmte_. We are using the version _3.2.0_ but there's a version _6.0.0_ from february 2020. Maybe this has been solved.

dominik-cnx:
> What about updating _jmte_. We are using the version _3.2.0_ but there's a version _6.0.0_ from february 2020. Maybe this has been solved.

The master branch in the repo mentioned in the description has the issue. Can you find a newer version in some other repo? I can easily see if this was fixed in the meantime.

pom.xml
dominik-cnx: Right, I'll look into this.

mcweba:
> > What about updating _jmte_. We are using the version _3.2.0_ but there's a version _6.0.0_ from february 2020. Maybe this has been solved.
> 
> The master branch in the repo mentioned in the description has the issue. Can you find a newer version in some other repo? I can easily see if this was fixed in the meantime.

I looked in the pom.xml of my local maven repository and found that the GitHub repository is [https://github.com/DJCordhose/jmte](https://github.com/DJCordhose/jmte) not the _HubSpot_ one. At least there the version _5.0.0_ is visible

mcweba:
The mentioned repo looks more active. Maybe we could open an issue so they are aware of the problem.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
```diff
@@ -211,11 +211,17 @@ public static String replaceWildcardConfigs(String contentWithWildcards,
...
      */
-    private static String replacePlainStringWildcardConfigs(String contentWithWildcards,
+    private static synchronized String replacePlainStringWildcardConfigs(String contentWithWildcards,
         Map<String, Object> properties) {
         if (properties == null || contentWithWildcards == null) {
             return contentWithWildcards;
         }
+        /*
+         * Note that this Engine is meant to be thread safe but it actually isn't because the TokenStream used by
+         * the Engine uses a MiniParser which in turn is not thread, see:
```
hiddenalpha, line 221: > `which in turn is not thread`

Typo? `thread` -> `thread save`

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
```diff
@@ -211,11 +211,17 @@ public static String replaceWildcardConfigs(String contentWithWildcards,
      * @return the String with replaced wildcard values. Returns the input String when input String or
      * properties are <code>null</code>
      */
-    private static String replacePlainStringWildcardConfigs(String contentWithWildcards,
+    private static synchronized String replacePlainStringWildcardConfigs(String contentWithWildcards,
```
hiddenalpha, line 214: Suggest to only synchronize the actual usage of the engine and not the whole method.

Synchronizing the whole method and then running into the early-return case (*props==null || conts==null*) 99% of the time would cost lots of performance due to useless locking.

Therefore I suggest to use somewhat like (*pseudCode*):
```java
if( fooOrBarIsNull ){
    return bar;
}
synchronized( StringUtils.class ){
    Engine engine = new Engine();
    return wildcardReplacementEngineIgnition(engine, contentWithWildcards, properties);
}
```

-> Same for the other method.

dominik-cnx:
> The mentioned repo looks more active. Maybe we could open an issue so they are aware of the problem.

Yes, I will do this.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
dominik-cnx: yep, will fix

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/StringUtils.java
dominik-cnx: OK, yes, makes sense, done

hiddenalpha (approved):

mcweba (approved):

PR #412, opened 2021-12-15
Issue400 schema location

mcweba:


gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
```diff
@@ -31,43 +31,42 @@ public DefaultCacheDataFetcher(HttpClient httpClient) {
...
+        );
 
-        cReq.setTimeout(requestTimeoutMs);
-        cReq.headers().setAll(requestHeaders);
-        cReq.headers().set("Accept", "application/json");
-        cReq.headers().set(SELF_REQUEST_HEADER, "true");
-        cReq.setChunked(true);
-        cReq.end();
+        fetchRequest.putHeader("Accept", "application/json");
+        fetchRequest.putHeader(SELF_REQUEST_HEADER, "true");
```
hiddenalpha, line 67: Any reason to set those headers after creating the request? Looks as the request creation method already takes headers. Are those the same?

gateleen-validation/src/main/java/org/swisspush/gateleen/validation/ValidationHandler.java
```diff
@@ -180,4 +186,28 @@ public void setFailOnError(boolean failOnError) {
...
+            }
+
+            String keepInMemoryStr = validationResource.get(ValidationResource.SCHEMA_KEEP_INMEMORY_PROPERTY);
+
+            Integer keepInMemory = null;
+            if(keepInMemoryStr != null) {
+                try {
+                    keepInMemory = Integer.parseInt(keepInMemoryStr);
+                } catch (NumberFormatException ex) {
+                    log.warn("Property 'keepInMemory' is not a number but " + keepInMemoryStr, ex);
```
hiddenalpha, line 204: http://www.slf4j.org/faq.html#paramException

gateleen-validation/src/main/java/org/swisspush/gateleen/validation/Validator.java
```diff
@@ -123,65 +157,71 @@ public static ValidationResult validateStatic(Buffer dataToBeValidated, String s
...
+    }
+
+    private static void performValidation(String dataString, JsonObject data, Logger log, String base, Buffer jsonBuffer,
+                                          String type, String path, Handler<ValidationResult> callback) {
+        if (SCHEMA_DECLARATION.equals(data.getString("$schema"))) {
             JsonSchema schema;
             try {
-                schema = JsonSchemaFactory.getInstance().getSchema(dataString);
+                schema = JsonSchemaFactory.getInstance(SpecVersion.VersionFlag.V4).getSchema(dataString);
             } catch (Exception e) {
```
hiddenalpha, line 210: [Pokémon Exception Handling](https://blog.codinghorror.com/new-programming-jargon/#2)

Would it be enough to catch `RuntimeException` or similar?

gateleen-validation/src/test/java/org/swisspush/gateleen/validation/ValidatorTest.java
```diff
@@ -61,20 +69,12 @@ public void testValidationWithVariables(TestContext context){
...
-                return "/foo/mediadata/v1/specials/03";
-            }
-        }
+        CustomHttpServerRequest gETValidationResourceRequest = new CustomHttpServerRequest(HttpMethod.GET, "/foo/mediadata/v1/specials/03");
 
         String type = "GET/out";
         Buffer jsonBuffer = Buffer.buffer();
         final SettableFuture<String> future = SettableFuture.create();
-        validator.validate(new GETValidationResourceRequest(), type, jsonBuffer, validationResult -> {
+        validator.validate(gETValidationResourceRequest, type, jsonBuffer, validationResult -> {
```
hiddenalpha, line 77:
```java
gETValidationResourceRequest
^^^^--- Nothing wrong. uPpeR/loWeR case just looks funny (aka counter-intuitive) :)
```
Maybe the classname should start with `GetVa...` instead `GETVa...`? (AFAIK `GET` is not an acronym)

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
mcweba: Yes those are the same headers. There is no particular reason to not already add those headers to the exsiting MultiMap. In this case, the intention was to show that those are additional headers for the client request.

gateleen-validation/src/main/java/org/swisspush/gateleen/validation/Validator.java
mcweba: Good point. I change it to `RuntimeException` because looking at the `JsonSchemaFactory` class we could get `IllegalArgumentException` and `JsonSchemaException`

mcweba: There are lots of places in gateleen were we do catch `Exception`. This should be targetet too.

lbovet (approved):

PR #425, opened 2022-02-15
allow to configure the HttpClientOptions.DEFAULT_KEEP_ALIVE_TIMEOUT on routes through system property

dominik-cnx:
As per title. With this PR we allow to configure the vert.x property. If not configured the default kicks in as before.

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
@@ -33,6 +33,9 @@ public class Route {
     private static final boolean CLIENT_DEFAULT_EXPAND_IN_STORAGE = false;
     private static final int CLIENT_DEFAULT_LOG_EXPIRY = 4 * 3600;
 
+    private static final String HTTP_CONNECTION_KEEP_ALIVE_TIMEOUT = "org.swisspush.gateleen.routing.rule.http.connection.keep.alive.timeout";
+    private static int keepAliveTimeout;
+
     private static final Pattern URL_PARSE_PATTERN = Pattern.compile("^(?<scheme>https?)://(?<host>[^/:]+)(:(?<port>[0-9]+))?(?<path>/.*)$");
     private static final Logger LOG = LoggerFactory.getLogger(Route.class);
     private static final Logger CLEANUP_LOGGER = LoggerFactory.getLogger(Route.class.getName() + "Cleanup");
@@ -50,6 +53,15 @@ public class Route {
     private HttpClient client;
     private Forwarder forwarder;
     private HttpClient selfClient;
+
+    static {
+        String keepAliveTimeoutProperty =  System.getProperty(HTTP_CONNECTION_KEEP_ALIVE_TIMEOUT);
+        if (keepAliveTimeoutProperty != null) {
+            keepAliveTimeout = Integer.parseInt(keepAliveTimeoutProperty);
+        } else {
+            keepAliveTimeout = HttpClientOptions.DEFAULT_KEEP_ALIVE_TIMEOUT;
+        }
+    }
 
     /**
      * Creates a new instance of a Route.
@@ -98,6 +110,7 @@ public class Route {
 
         rule.setTimeout(1000 * CLIENT_DEFAULT_TIMEOUT_SEC);
         rule.setKeepAlive(CLIENT_DEFAULT_KEEP_ALIVE);
+        rule.setKeepAliveTimeout(keepAliveTimeout);
         rule.setExpandOnBackend(CLIENT_DEFAULT_EXPAND_ON_BACKEND);
         rule.setStorageExpand(CLIENT_DEFAULT_EXPAND_IN_STORAGE);
         rule.setLogExpiry(CLIENT_DEFAULT_LOG_EXPIRY);
--- a/gateleen-kafka/src/test/java/org/swisspush/gateleen/kafka/KafkaHandlerTest.java
+++ b/gateleen-kafka/src/test/java/org/swisspush/gateleen/kafka/KafkaHandlerTest.java
@@ -35,6 +35,7 @@ import java.util.regex.Pattern;
 import static com.jayway.awaitility.Awaitility.await;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.hamcrest.Matchers.equalTo;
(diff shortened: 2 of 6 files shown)


gateleen-kafka/src/test/java/org/swisspush/gateleen/kafka/KafkaHandlerTest.java
```diff
@@ -128,15 +129,15 @@ public void initWithWildcardConfigResource(TestContext context) {
         context.assertFalse(handler.isInitialized());
 
         handler.initialize().setHandler(event -> {
-            verify(repository, times(2)).closeAll();
-
+            // depending whether resourceChanged fires or not we get one or two invocations
```
dominik-cnx, line 132: not related to the PR, the test sometimes fails on travis
@geekrumper can you kindly have a look if this makes sense / propose changes as needed, thx.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
```diff
@@ -51,6 +54,15 @@
     private Forwarder forwarder;
     private HttpClient selfClient;
 
+    static {
+        String keepAliveTimeoutProperty =  System.getProperty(HTTP_CONNECTION_KEEP_ALIVE_TIMEOUT);
```
mcweba, line 58: To be a bit more consistent in the usage of system properties you could introduce the `Map<String, Objec>` properties map. This is already used in several other Gateleen classes like Handlers.

For example in Crush, we also have the following code in the `MainVerticle`, which allows to define such properties based on environment:
```java
final Map<String, Object> props;
try {
	props = new ResourcePropertySource("classpath:/" + SERVER_NAME + "/config/environments/default-env.properties").getSource();

	Resource envResource = new ClassPathResource("/" + SERVER_NAME + "/config/environments/" + env + "-env.properties");
	if (envResource.exists()) {
		props.putAll(new ResourcePropertySource(envResource).getSource());
		log.info("Merging environment config from classpath for " + env);
	} else {
		log.info("No environment config in classpath for " + env);
	}

	String externalConfig = System.getProperty(CONFIG_DIR_SYSTEM_PROPERTY) + "/nemo/config.properties";

	Resource externalConfigResource = new FileSystemResource(externalConfig);

	if (externalConfigResource.exists()) {
		props.putAll(new ResourcePropertySource(externalConfigResource).getSource());
		log.info("Merging external config " + externalConfig);
	} else {
		log.info("No external config found under " + externalConfig);
	}
} catch (IOException e) {
	throw new RuntimeException(e);
}
```

By defining the `keepAliveTimeout` here, it would be used for all routes? Wouldn't it be better to allow the `keepAliveTimeout` in the configuration resource of routes and rules? So we could define it just for the routes/rules we want. With the global approach we could even define a global keepAlive and override it with specific configuration on route and rule basis.

I would suggest to adopt the json schema for routes and rules to allow the configuration of a `keepAliveTimeout` value

gateleen-kafka/src/test/java/org/swisspush/gateleen/kafka/KafkaHandlerTest.java
geekrumper: @dominik-cnx I think this is okay, if we can't guarantee that it always triggers two times.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
dominik-cnx: In our case we can easily define system properties on a per stage base outside of some gateleen mechanism. What we have here is totally sufficient.

We actually want to use the keepAliveTimeout for dynamically created routes, we do not want to set in on per statically configured route in the rules resource.

mcweba: You added the possibility to define the keepAliveTimeout on routing rules but it is only usable for dynamically created routes. Even if you don't need it right now, I would still add it also for statically configured routes/rules. Besides that, it's a small effort and others could use it someday. I also think its not correct do have configurations on rules which you cannot use everywhere.

dominik-cnx: Hmm, yeah, OK. HTTP/1.1 will probably stay with us a little bit longer, so there is a chance. Will add it.

mcweba (approved):

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
```diff
@@ -24,6 +24,7 @@
     private int poolSize;
     private int maxWaitQueueSize;
     private boolean keepAlive;
+    private int keepAliveTimeout;
```
hiddenalpha, line 27: Same as (*TODO insert link*)

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
```diff
@@ -57,6 +57,11 @@
 					"type": "integer",
 					"default": 30
 				},
+				"keepAliveTimeout": {
+                    "description": "The keep alive timeout for HTTP/1.1 connections.",
```
hiddenalpha, line 61: Suggest to mention unit in description.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/Route.java
```diff
@@ -33,6 +33,9 @@
     private static final boolean CLIENT_DEFAULT_EXPAND_IN_STORAGE = false;
     private static final int CLIENT_DEFAULT_LOG_EXPIRY = 4 * 3600;
 
+    private static final String HTTP_CONNECTION_KEEP_ALIVE_TIMEOUT = "org.swisspush.gateleen.routing.rule.http.connection.keep.alive.timeout";
+    private static int keepAliveTimeout;
```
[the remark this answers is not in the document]
dominik-cnx: I did consider this but decided against it because this pattern is not in use so far within gateleen.

gateleen-routing/src/main/resources/gateleen_routing_schema_routing_rules
dominik-cnx: Missed this right, in the meantime we have merged. I'll try to remember this for my next PR.

PR #426, opened 2022-02-24
Updated to vertx 4.2.1

dominik-cnx:
Uses the friends that are already on 4.2.1.

Bumped the version from 1.2.x to 1.3.0-SNAPSHOT because this is major change.


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/DeferCloseHttpClient.java
```diff
@@ -39,54 +34,13 @@ public DeferCloseHttpClient(Vertx vertx, HttpClient delegate) {
...
     @Override
-    public HttpClientRequest request(HttpMethod method, int port, String host, String requestURI, Handler<HttpClientResponse> responseHandler) {
+    public void request(HttpMethod method, int port, String host, String requestURI, Handler<AsyncResult<HttpClientRequest>> handler) {
         logger.debug("({}:{}).request({}, \"{}\")", host, port, method, requestURI);
         countOfRequestsInProgress += 1;
         logger.debug("Pending request count: {}", countOfRequestsInProgress);
-
-        // Delegate to the same method on the delegate. But install our own handler which
-        // allows us to intercept the response.
-        HttpClientRequest request = delegate.request(method, port, host, requestURI, upstreamRsp -> {
```
dominik-cnx, line 50: @hiddenalpha do we need to migrate this? Or do you think that this is no longer needed with the new vert.x version?

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
```diff
@@ -44,48 +45,59 @@ public DefaultCacheDataFetcher(ClientRequestCreator clientRequestCreator, String
...
                 HttpMethod.GET,
                 requestUri,
                 requestHeaders,
-                requestTimeoutMs,
-                cRes -> cRes.bodyHandler(data -> {
+                requestTimeoutMs, event -> {
+                    log.warn("Got an error while fetching cache data", event);
+                    promise.complete(Result.err(StatusCode.INTERNAL_SERVER_ERROR));
+                }).onComplete(event -> {
+            if (event.failed()) {
```
hiddenalpha, line 60: Why this strange indentation?

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
```diff
@@ -44,48 +45,59 @@ public DefaultCacheDataFetcher(ClientRequestCreator clientRequestCreator, String
...
                         }
 
-                        future.complete(Result.ok(data));
+                        promise.complete(Result.ok(data));
                     } else {
                         StatusCode statusCode = StatusCode.fromCode(cRes.statusCode());
-                        if(statusCode == null) {
-                            log.error("Got unknown status code {} while fetching cache data. Using 500 Internal Server Error " +
+                        if (statusCode == null) {
+                            log.error("Got unkown status code {} while fetching cache data. Using 500 Internal Server Error " +
```
hiddenalpha, line 86: Why re-introducing a typo?

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/storage/CacheStorage.java
```diff
@@ -1,6 +1,7 @@
 package org.swisspush.gateleen.cache.storage;
 
 import io.vertx.core.Future;
+import io.vertx.core.Promise;
```
hiddenalpha, line 4: Import looks unused as its the only change in the file

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/storage/RedisCacheStorage.java
```diff
@@ -77,87 +77,93 @@ private long lockExpiry(long taskInterval) {
...
                 log.error(message);
-                future.fail(message);
+                promise.fail(message);
             } else {
-                JsonArray array = reply.result();
+                JsonArray array = new JsonArray();
+                reply.result().stream().parallel()
+                        .forEach(e -> {
+                            synchronized (array) {
+                                array.add(e);
```
hiddenalpha, line 146: **WARN**: WTH?!? - Using a parallel stream and then using a synchronized block inside? -> That looks pretty much like a lot of nonsense.

hiddenalpha (changes requested):

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/DeferCloseHttpClient.java
hiddenalpha: Removing this defeats the purpose of the whole class (*aka breaks it*). I bet we still need this. Because without the reference counting, we have the old behavior if the class wouldn't exist at all. And therefore we will re-introduce the bug we solved with it.

dominik-cnx: Kind of expected that we probably continue to need it. Would you be OK if we merge now to develop
- with this open issue
- and consider the release not fit for production
- and follow up right away
?

The reason I ask is this: I would like to integrate the release in our project and run all automated integration tests. In order to do this we need to stage to maven central (which means to release, we normally do this through develop / master). Technically I can also release from this branch I believe. An other advantage to have it arrive in develop soon  (beside the ability to release as usual) is to avoid future merge conflicts.

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
dominik-cnx: The default auto-formatter in Intellij formats like this.

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/fetch/DefaultCacheDataFetcher.java
dominik-cnx: Probably a merge issue, will fix

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/storage/CacheStorage.java
dominik-cnx: Yep, thx, will fix

gateleen-cache/src/main/java/org/swisspush/gateleen/cache/storage/RedisCacheStorage.java
dominik-cnx: Yep, weird, will have a look

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/DeferCloseHttpClient.java
dominik-cnx: Discussed, we add a TODO, maybe something like

TODO: Migrate the original implementation

and then also look into it right away. An option is probably also to introduce this DeferCloseHttpClient in vert.x itself.

PR #444, opened 2022-03-23
Use chrome headless instead of phantom js

dominik-cnx:
As per title, we now use chrome for karma testing.

### DIFF
--- a/gateleen-hook-js/pom.xml
+++ b/gateleen-hook-js/pom.xml
@@ -51,6 +51,23 @@
                             </resources>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>prepare-js-for-tests</id>
+                        <phase>compile</phase>
+                        <goals>
+                            <goal>copy-resources</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>
+                                ${project.basedir}/target/dist/js
+                            </outputDirectory>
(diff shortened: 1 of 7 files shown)


gateleen-hook-js/src/test/javascript/hook/gateleen-hook.test.js
```diff
@@ -72,25 +72,6 @@ describe('hook', function(){
       $httpBackend.verifyNoOutstandingExpectation();
       $httpBackend.verifyNoOutstandingRequest();
     });
-    it('should create the hook with HTTP PUT from complete URL', function(){
-      $httpBackend.expectPUT('http://server/context/path/_hooks/listeners/http/gateleen-hook-js-0123',
```
dominik-cnx, line 76: The current production code doesn't support this case and it doesn't seem to be in use/required. The tests were not executed on Travis, hence we didn't notice.

gateleen-hook-js/pom.xml
```diff
@@ -98,7 +115,36 @@
                             <arguments>build</arguments>
                         </configuration>
                     </execution>
+                    <execution>
+                        <id>grunt test</id>
```
dominik-cnx, line 119: This is needed to actually execute the test upon Travis build

PR #447, opened 2022-03-24
Refactored RedisMonitor since redisAPI does not return a json anymore…

mcweba:
… but plain text

### DIFF
--- a/gateleen-monitoring/src/main/java/org/swisspush/gateleen/monitoring/RedisMonitor.java
+++ b/gateleen-monitoring/src/main/java/org/swisspush/gateleen/monitoring/RedisMonitor.java
@@ -115,8 +115,12 @@ public class RedisMonitor {
                     String[] pairs = val.split(",");
                     for (String pair : pairs) {
                         String[] tokens = pair.split("=");
-                        value = Long.parseLong(tokens[1]);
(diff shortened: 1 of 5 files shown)


gateleen-monitoring/src/main/java/org/swisspush/gateleen/monitoring/RedisMonitor.java
```diff
@@ -115,8 +115,12 @@ public String apply(@Nullable String input) {
                     String[] pairs = val.split(",");
                     for (String pair : pairs) {
                         String[] tokens = pair.split("=");
-                        value = Long.parseLong(tokens[1]);
-                        publisher.publishMetric("keyspace." + key + "." + tokens[0], value);
+                        if(tokens.length == 2) {
+                            value = Long.parseLong(tokens[1]);
+                            publisher.publishMetric("keyspace." + key + "." + tokens[0], value);
+                        } else {
+                            log.warn("Invalid keyspace property. Will be ignored");
```
hiddenalpha, line 122: Any reason to not log any details? For example I think contents of "*pair*" probably would be very interesting for bug-hunting. Like:
```java
log.warn("Invalid keyspace property. Will be ignored: {}", pair)
```

PR #456, opened 2022-04-07
Avoid exception due to multiple Content-Length header values

dominik-cnx:
We had this here in our project for a certain forwarding case

2022-03-31 09:12:18,517 test blah ERROR ConnectionBase - Multiple Content-Length values found: [158, 158]
2022-03-31 09:12:18,517 test blah ERROR EnqueuePatrolManager - Failed to fetch queueMetadata.
java.lang.IllegalArgumentException: Multiple Content-Length values found: [158, 158]
    at io.netty.handler.codec.http.HttpUtil.normalizeAndGetContentLength(HttpUtil.java:599)

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -1,15 +1,23 @@
 package org.swisspush.gateleen.core.util;
 
 import io.vertx.core.MultiMap;
+import io.vertx.core.http.HttpServerResponse;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 
 public class HttpHeaderUtil {
+    private static final Logger LOG = LoggerFactory.getLogger(HttpHeaderUtil.class);
 
     /**
      * <p>Removes headers which MUST NOT be forwarded by proxies.</p>
@@ -19,12 +27,10 @@ public class HttpHeaderUtil {
      *     <li>This MAY returns the same (modified) instance as passed.</li>
      * </ul>
      *
+     * @param headers The headers to check.
      * @see <a href="https://tools.ietf.org/html/rfc2616#section-14.10">RFC 2616 section 14.10</a>
-     *
-     * @param headers
-     *      The headers to check.
      */
-    public static <T extends MultiMap> T removeNonForwardHeaders( T headers ) {
+    public static <T extends MultiMap> T removeNonForwardHeaders(T headers) {
         final String CONNECTION = HttpRequestHeader.CONNECTION.getName();
 
         // Remove all headers named by connection-token.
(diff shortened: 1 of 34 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -61,18 +68,43 @@
...
+    /**
+     * Merges headers, makes sure that only one value of header {@link HttpRequestHeader#CONTENT_LENGTH} ends up in the result.
+     */
+    public static void mergeHeaders(@Nonnull MultiMap destination, @Nonnull MultiMap source, @Nullable String context) {
+        String destinationContentLength = "";
+        String sourceContentLength = "";
+
+        if (HttpRequestHeader.containsHeader(destination, HttpRequestHeader.CONTENT_LENGTH)) {
+            destinationContentLength = destination.get(HttpRequestHeader.CONTENT_LENGTH.getName());
+        }
```
hiddenalpha, line 94: (*trivial*) [Doc of MultiMap.get(String)](https://vertx.io/docs/apidocs/io/vertx/core/MultiMap.html#get-java.lang.String-) says:
> Returns: The first header value **or null if there is no such entry**

I guess could also be written without the `contains` check:
```java
    destinationContentLength = destination.get(HttpRequestHeader.CONTENT_LENGTH.getName());
    if( destinationContentLength == null ){
        destinationContentLength = "";
    }
```
Just an idea. I think its not important how we do it.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -61,18 +68,43 @@
...
+        if (HttpRequestHeader.containsHeader(source, HttpRequestHeader.CONTENT_LENGTH)) {
+            sourceContentLength = source.get(HttpRequestHeader.CONTENT_LENGTH.getName());
+        }
+        if (destinationContentLength.isEmpty() || sourceContentLength.isEmpty()) {
+            destination.addAll(source);
+        } else {
+            if (!destinationContentLength.equals(sourceContentLength)) {
+                LOG.warn("Content-Length values do not match {} != {} for request to url {}",
+                        destinationContentLength, sourceContentLength, context);
```
hiddenalpha, line 103: Not sure if this is a good idea to just continue in this case. I think we should fail if the values differ (eg throw or whatever).

Continuing here potentially leads to unhealthy out-of-sync http connections.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -61,18 +68,43 @@
...
                 return true;
             }
         }
         return false;
     }
 
+    /**
+     * Merges headers, makes sure that only one value of header {@link HttpRequestHeader#CONTENT_LENGTH} ends up in the result.
+     */
+    public static void mergeHeaders(@Nonnull MultiMap destination, @Nonnull MultiMap source, @Nullable String context) {
```
hiddenalpha, line 88: I think would be helpful to document at least the parameter "*context*". I think something like the following could clarify things for callers:
> @param context  Only used for log/error messages. Does not impact how we merge.

dominik-cnx: Right, will improve, thx.

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
dominik-cnx: Yes, but the readabiltity is better with "contains" I would say. Implemention wise we end up in a get() in both cases

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
dominik-cnx: Basically agree. It would mean to implement the error handling for some 10 usages of this merge method. Hmm, not sure if we can justify the effort. But I changed the log into an error to indicate that there is clearly something not as it should be.

PR #460, opened 2022-04-22
Remove all duplicated headers

dominik-cnx:
With this PR we provide means to remove all duplicated headers, so far we had special treatment for header "Content-Length" only.

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
@@ -78,22 +78,21 @@ public class HttpHeaderUtil {
     }
 
     /**
-     * Merges headers, makes sure that only one value of header ends up in the result.
+     * Merges headers, makes sure that only one value of all header ends up in the result.
      *
      * @param context optional context information to be used for logging purposes, not used for the actual merge
+     *                Note: This is not 100% in line with https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2 to be further looked into.
      */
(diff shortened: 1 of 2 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
```diff
@@ -83,30 +78,28 @@ public static <T extends MultiMap> boolean hasMatchingHeader(@Nonnull T headers,
     }
 
     /**
-     * Merges headers, makes sure that only one value of header {@link HttpRequestHeader#CONTENT_LENGTH} ends up in the result.
+     * Merges headers, makes sure that only one value of all headers ends up in the destination MultiMap.
+     *
+     * Note: This is not 100% in line with https://www.rfc-editor.org/rfc/rfc7230#section-3.2.2 to be further looked into.
```
dominik-cnx, line 83: I'm not 100% convinced yet and hope to find the root cause why we actually end up in situations where we have header duplication. However, we are currently blocked in our project due to this issue.

Kusig: Is there a specific pattern or header which is duplicated? Some expamples? This might give a sign to the root of the problem.

Kusig (approved):

gateleen-core/src/main/java/org/swisspush/gateleen/core/util/HttpHeaderUtil.java
dominik-cnx: image

PR #464, opened 2022-05-17
Forwarder hangs in some cases if queue is involved

dominik-cnx:
With this PR we fix a case where a request hangs and then timeouts. I wasn't able to write an integration test for this particular test within Gateleen itself but I do have one in my project. I'm happy to provide such a test in Gateleen itself if required, let me know. It will take me some time though.

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/BufferBridge.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/http/BufferBridge.java
@@ -21,7 +21,7 @@ public class BufferBridge {
     private Handler<Buffer> bodyHandler;
     private Handler<Throwable> exceptionHandler;
     private Queue<Buffer> queue = new LinkedList<>();
-    private Buffer body;
+    private Buffer body = Buffer.buffer();
     private boolean ended = false;
     private Vertx vertx;
     private final Logger log = LoggerFactory.getLogger(BufferBridge.class);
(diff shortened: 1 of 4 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/http/BufferBridge.java
```diff
@@ -67,11 +70,14 @@ protected void doWrite(Buffer chunk) {
 
     protected Future<Void> doEnd() {
         ended = true;
-        if(endHandler != null && queue.isEmpty()) {
+        if (bodyHandler != null) {
```
dominik-cnx, line 73: Follows the same approach as in https://github.com/eclipse-vertx/vert.x/blob/4.2.1/src/main/java/io/vertx/core/http/impl/HttpEventHandler.java#L79-L82

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/DeferCloseHttpClient.java
```diff
@@ -43,9 +43,16 @@ public void request(HttpMethod method, int port, String host, String requestURI,
         countOfRequestsInProgress += 1;
         logger.debug("Pending request count: {}", countOfRequestsInProgress);
         delegate.request(method, port, host, requestURI).onComplete(asyncRequestResult -> {
+            if (asyncRequestResult.failed()) {
```
dominik-cnx, line 46: Not related to the PR, prevents a NPE

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -312,8 +312,7 @@ public Future<Void> write(Buffer data) {
                                 firstBuffer = false;
                                 cReq.setChunked(true);
                             }
-                            cReq.write(data);
-                            return Future.succeededFuture();
+                            return cReq.write(data);
```
dominik-cnx, line 315: Not related to the PR

mcweba (approved):

PR #488, opened 2022-11-28
Consider the number of route instances in a cluster when applying con…

dominik-cnx:
…figured connection pool sizes.

So far the connection pool size configured on static routes or route hooks, e.g.
 
  "/to/my-backend/(.*)": {
    "url": "http://host.domain.com:7022/$1",
    "connectionPoolSize": 7,
  },

was applied on a per cluster node based AND also on a per router instance base if there were multiple instances. In a case where the number of nodes in a cluster varies (e.g. scaling up is performed due to high load or similar) then the number of concurrent active connections is higher compared to what is configured. This can be confusing and/or actually lead to situations where the backend starts to refuse connection attempts.

With this PR we change this behavior and divide the configured connection pool size by the number of router instances (a so called "route multiplier") which can be passed in and updated at runtime through the event bus. To actually pass in the "route multiplier" is optional, if not passed in things remain unchanged.

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -1533,15 +1533,16 @@ public class HookHandler implements LoggableResource {
 
         // Configure connection pool size
         Integer originalPoolSize = jsonHook.getInteger(HttpHook.CONNECTION_POOL_SIZE_PROPERTY_NAME);
-        int appliedPoolSize = 0;
+        int appliedPoolSize;
         if (originalPoolSize != null) {
             appliedPoolSize = Math.floorDiv(originalPoolSize, routeMultiplier);
             if (appliedPoolSize < 1) {
                 appliedPoolSize = originalPoolSize;
             }
+            log.debug("Original pool size is {}, applied size is {}", originalPoolSize, appliedPoolSize);
+            hook.setConnectionPoolSize(appliedPoolSize);
         }
-        log.debug("Original pool size is {}, applied size is {}", originalPoolSize, appliedPoolSize);
-        hook.setConnectionPoolSize(appliedPoolSize);
+
 
         hook.setMaxWaitQueueSize(jsonHook.getInteger(HttpHook.CONNECTION_MAX_WAIT_QUEUE_SIZE_PROPERTY_NAME));
         // Configure request timeout
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -445,9 +445,13 @@ public class HookHandler implements LoggableResource {
      * @param readyHandler - the ready handler
      */
     private void registerRouteMultiplierChangeHandler(Handler<Void> readyHandler) {
-        vertx.eventBus().consumer(Router.ROUTE_MULTIPLIER_ADRESS, (Handler<Message<String>>) event -> {
-            log.warn("Update router's pool size multiplier: {}", (event.body() == null ? "<null>" : event.body()));
-            routeMultiplier = Integer.parseInt(event.body());
+        vertx.eventBus().consumer(Router.ROUTE_MULTIPLIER_ADDRESS, (Handler<Message<String>>) event -> {
+            log.info("Updating route multiplier: {}", (event.body() == null ? "<null>" : event.body()));
+            try {
+                routeMultiplier = Integer.parseInt(event.body());
+            } catch (NumberFormatException e) {
+                log.info("failed to parse route multiplier: {}", event.body(), e);
+            }
         });
         // method done / no async processing pending
         readyHandler.handle(null);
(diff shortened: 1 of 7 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -423,6 +439,20 @@ private void registerRouteRegistrationHandler(Handler<Void> readyHandler) {
...
 
+    /**
+     * Update all registered router's pool size multiplier
+     *
+     * @param readyHandler - the ready handler
+     */
+    private void registerRouteMultiplierChangeHandler(Handler<Void> readyHandler) {
+        vertx.eventBus().consumer(Router.ROUTE_MULTIPLIER_ADRESS, (Handler<Message<String>>) event -> {
+            log.warn("Update router's pool size multiplier: {}", (event.body() == null ? "<null>" : event.body()));
+            routeMultiplier = Integer.parseInt(event.body());
```
mcweba, line 450: I think we should catch exceptions here like _NumberFormatException_. What happens when we get a non-parsable String?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -423,6 +439,20 @@ private void registerRouteRegistrationHandler(Handler<Void> readyHandler) {
         readyHandler.handle(null);
     }
 
+    /**
+     * Update all registered router's pool size multiplier
+     *
+     * @param readyHandler - the ready handler
+     */
+    private void registerRouteMultiplierChangeHandler(Handler<Void> readyHandler) {
+        vertx.eventBus().consumer(Router.ROUTE_MULTIPLIER_ADRESS, (Handler<Message<String>>) event -> {
```
mcweba, line 448: Where are the messages to _Router.ROUTE_MULTIPLIER_ADRESS_ created? I only find consumers.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
```diff
@@ -43,9 +43,11 @@ public class Router implements Refreshable, LoggableResource, ConfigurationResou
      * How long to let the http clients live before closing them after a re-configuration
      */
     private static final int GRACE_PERIOD = 30000;
+    public static final int DEFAULT_ROUTER_MULTIPLIER = 1;
     public static final String ROUTER_STATE_MAP = "router_state_map";
     public static final String ROUTER_BROKEN_KEY = "router_broken";
     public static final String REQUEST_HOPS_LIMIT_PROPERTY = "request.hops.limit";
+    public static final String ROUTE_MULTIPLIER_ADRESS = "gateleen.route-multiplier";
```
mcweba, line 50: Typo in _ROUTE_MULTIPLIER_ADRESS_

mcweba (changes requested):

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
dominik-cnx: OK, will add

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
dominik-cnx: I have a utility class in my project that repeatedly determines the number of nodes currently up and running


    public static int getRouteMultiplier(Vertx vertx){
        int numberOfClusterNodes = 1;
        ClusterManager clusterManager = ((VertxImpl)vertx).getClusterManager();
        if (clusterManager != null){
            numberOfClusterNodes = clusterManager.getNodes().size();
        }
        return Props.getNumberOfMainVerticleInstances() * numberOfClusterNodes;
    }
    
  
    Note that in my case the number of Verticles that create a Router is fixed and doesn't change at runtime.
    
    
It then publishes like this if things change

    private static void startRouteMultiplierPublishing(Vertx vertx){
        vertx.setPeriodic(10 * 1000, event -> {
            final EventBus eb = vertx.eventBus();
            int newRouteMultiplier = MainVerticle.getRouteMultiplier(vertx);
            if (previousRouteMultiplier != newRouteMultiplier) {
                eb.publish(Router.ROUTE_MULTIPLIER_ADRESS, String.valueOf(MainVerticle.getRouteMultiplier(vertx)));
                previousRouteMultiplier = newRouteMultiplier;
            }
        });
    }

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
dominik-cnx: Thx, fixed

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
mcweba: OK. I was looking for an API endpoint or a configuration resource. But this could be added "manually" in Crush/Houston if required.

mcweba (approved):

PR #494, opened 2023-01-19
Sync log levels across cluster nodes

dominik-cnx:
As per title, the idea is that log level changes become active on all nodes in the cluster (and not only on the node that receives the change request)


mcweba (commented):
I think we should let the user somehow decide whether to apply the changes to all cluster nodes. A use case for not syncing the log levels could be to apply the change only to one node because of heavy traffic or because just one node has problems that require investigation.

dominik-cnx:
OK, interesting, in our project it would be a bit difficult to control on which node the change will be performed, the user doesn't really "see" the individual (and also the logs of all nodes end up in Kibana in our case).

dominik-cnx:
I'll see if I can introduce an additional boolean parameter named applyToAllClusterNodes or similar.

dominik-cnx:
In the meantime we have implemented a solution on our side and no longer need this change.

PR #495, opened 2023-01-25
Removed duplication of request header holding the unique request id.

mcweba:
The second header (x-rp-unique-id) was introduced back in 2016. However, I did not find a place where this duplicated header is actually used. To reduce confusion I removed this header.


dominik-cnx:
I can find references to x-rp-unique-id in our project but do believe that we can remove those. 

I wonder if we should change from

| x-request-id | unique id of the request when no x-rp-unique_id request header was provided

to

| x-request-id | unique id of the request, x-rp-unique_id if provided by the proxy or some gateleen generated id if the proxy doesn't provided the x-rp-unique_id.

The advantage would be that we always have some id in x-request-id

hiddenalpha:
IF I read code/comments around their occurrence it seems as this PR has chosen to go for the deprecated/old variant of the two. Reading the comments it sounds as the lodash variant is the older/deprecated one and the dash variant would be the newer one.

But that be said: I've no idea what those headers are for. Therefore I don't really care which one we remove ;)

hiddenalpha (approved):

PR #496, opened 2023-02-07
listeners of type "after" do not fire if a route hook is set on the same URL in some cases

dominik-cnx:
This is relevant for a case where we have a listener hook of type "after"

  PUT http://localhost:7013/**somehost/data/mytest**/_hooks/**listeners**/http/somehost/just-a-test
  {
    "methods": [
      "POST", 
      "PUT", 
      "DELETE"
    ],
    "destination": "/somehost/data/another-target", 
    "type": "**after**"
  }

and on the very same endpoint also a route hook

PUT http://localhost:7013/**somehost/data/mytest**/_hooks/**route**
  {
    "methods": [
      "POST", 
      "PUT", 
      "DELETE"
    ],
    "destination": "/somehost/data/mytestroutetarget"
  }  

and there is also a static route like this

  "/**somehost/data/mytest**/(.*)": {
    "path": "/somehost/server/return-http-error/503"
  },

With this PR the listener hook will be called in case that the request can be successfully forward with the route hook, before it wasn't called.

### DIFF
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -103,7 +103,7 @@ public class Forwarder extends AbstractForwarder {
 
     @Override
     public void handle(final RoutingContext ctx) {
-        handle(ctx.request(), null);
+        handle(ctx.request(), null, null);
     }
 
     /**
@@ -116,7 +116,7 @@ public class Forwarder extends AbstractForwarder {
      * @param bodyData - a buffer with the body data, null if the request
      *                 was not yet consumed
      */
-    public void handle(final HttpServerRequest req, final Buffer bodyData) {
+    public void handle(final HttpServerRequest req, final Buffer bodyData, @Nullable final Handler<Void> afterHandler) {
         final Logger log = RequestLoggerFactory.getLogger(Forwarder.class, req);
 
         if (handleHeadersFilter(req)) {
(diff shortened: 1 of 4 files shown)


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -475,6 +475,11 @@ private Handler<AsyncResult<HttpClientResponse>> getAsyncHttpClientResponseHandl
             cRes.endHandler(v -> {
                 try {
                     req.response().end();
+
+                    // if everything is fine, we call the after handler
+                    if (afterHandler != null && cRes.statusCode() == StatusCode.OK.getStatusCode()) {
```
mcweba, line 480: Are you sure that the status code **200 OK** is enough? What happens when a client request goes into a queue? Then you will get a **202 Accepted** as the answer. Maybe there are other cases like **304 Not Modified** which are also OK to continue afterward.

dominik-cnx: Good point, this was taken from

https://github.com/swisspost/gateleen/blob/13617adf3205fcc3badd110337bc6d67c68377ee/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java#L1203

so do we have the same problem there as well? Not sure if we have to prepare for a 304, we normally don't see this for a PUT / POST / DELETE?

mcweba: I agree that we have the same problem also in the HookHandler. **304 Not Modified** status codes are actually (at least in the gateleen environment) only used for GET requests where the Storage and EtagHandler are involved.

I would suggest checking for any **2xx** status codes. They should all be OK. What do you think?

dominik-cnx: OK, let's do this

PR #499, opened 2023-02-14
changed the poolsize evaluation if running in a cluster, simply round up

dominik-cnx:
The behavior change is here:

NOW:
Assert.assertEquals(1, RuleFactory.evaluatePoolSize(4, 5)); // configured 4, 5 cluster nodes
-> results in a total of 5 connection, close to what is configured

BEFORE:
Assert.assertEquals(4, RuleFactory.evaluatePoolSize(4, 5)); // configured 4, 5 cluster nodes
-> results in a total of 20 connection, higher than configured

The idea is to protect the third party / external system when running in a cluster configuration. In cases where small connection pool sizes are configured (smaller than the number of cluster nodes) we used the originalPoolSize. This resulted in a pool size of n x originalPoolSize which is possibly more then the external system can handle / against the idea of protection. With this PR we now end up with a connection pool size of 1 (per node) in such situations.


mcweba (approved):

mcweba (changes requested):
RuleFactoryTest fails

PR #503, opened 2023-02-27
Feature/oauth20 provider

mcweba:


gateleen-routing/README_routing.md
```diff
@@ -3,29 +3,60 @@
...
-### Usage
-To use the request hop validation feature, the _ConfigurationResourceManager_ and the path to the configuration resource have to be configured in the _Router_ by calling:
-
-```java
-router.enableRoutingConfiguration(configurationResourceManager, SERVER_ROOT + "/admin/v1/routing/config")
-```
-The routing configuration is a json resource defining the limit for the maximum request routings (hops) called **_request.hops.limit_**.
+### OAuth2 authentication
+To activate OAuth2 authentication for forwarding requests, a configuration with the required properties like _flowType_, _clientId_, _clientSecret_, etc. can be provided. See [vertx-auth-oauth2 documentation](https://vertx.io/docs/vertx-auth-oauth2/java/) for more
+information. The name of the configuration (_demo-config-1_ in the example) is used in the routing rule.
```
dominik-cnx, line 50: Should we then better use 'id' instead of 'name' to indicate that?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -149,24 +149,34 @@ public void handle(final HttpServerRequest req, final Buffer bodyData, @Nullable
...
-            });
-        } else {
-            handleRequest(req, bodyData, targetUri, log, null, afterHandler);
-        }
+
+        maybeAuthenticate(rule).onComplete(event -> {
+            if(event.failed()) {
+                req.resume();
+                log.error("Failed to authenticate request. Cause: {}", event.cause().getMessage());
+                respondError(req, StatusCode.INTERNAL_SERVER_ERROR);
```
dominik-cnx, line 157: What happens if access is not granted, we should then return some 401/3, no?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
```diff
@@ -221,6 +131,13 @@ public Router(Vertx vertx,
         this.httpClientFactory = httpClientFactory;
         this.doneHandlers = doneHandlers;
         this.routeMultiplier = routeMultiplier;
+        this.oAuthProvider = oAuthProvider;
+
+        if(oAuthProvider != null) {
+            this.oAuthStrategy = new OAuthStrategy(oAuthProvider);
+        }
+        this.basicAuthStrategy = new BasicAuthStrategy();
```
dominik-cnx, line 139: This means that we can have two Strategies active, plan A and plan B. Shouldn't it be more one or the other?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Container holding configuration values for the {@link Router}
+ *
+ * @author https://github.com/mcweba [Marc-Andre Weber]
+ */
+public class RouterConfiguration {
+
+    private final Integer requestHopsLimit;
+    private final Map<String, OAuthConfiguration> oAuthConfigurations;
```
dominik-cnx, line 16: Maybe indicate what the key is

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfigurationParser.java
```diff
@@ -0,0 +1,113 @@
...
+    private static final String REQUEST_HOPS_LIMIT_PROPERTY = "request.hops.limit";
+    private static final String AUTHCONFIGS_PROPERTY = "authConfigs";
+    private static final String FLOWTYPE_PROPERTY = "flowType";
+    private static final String CLIENTID_PROPERTY = "clientId";
+    private static final String CLIENTSECRET_PROPERTY = "clientSecret";
+    private static final String SITE_PROPERTY = "site";
+    private static final String TOKENPATH_PROPERTY = "tokenPath";
+    private static final String AUTHPATH_PROPERTY = "authPath";
+    private static final String SCOPES_PROPERTY = "scopes";
+    private static final String SUPPORTEDGRANTTYPES_PROPERTY = "supportedGrantTypes";
```
dominik-cnx, line 30: Trivial, we could add some extra _, eg.

 SUPPORTED_GRANTTYPES_PROPERTY 
AUTH_CONFIGS_PROPERTY

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfigurationParser.java
```diff
@@ -0,0 +1,113 @@
...
+        String replacedConfig;
+        JsonObject config;
+
+        try {
+            replacedConfig = StringUtils.replaceWildcardConfigs(configurationResourceBuffer.toString(UTF_8), properties);
+            config = new JsonObject(Buffer.buffer(replacedConfig));
+        } catch (Exception e) {
+            log.warn("Could not replace wildcards with environment properties for the router configuration " +
+                            "due to following reason: {}", e.getMessage());
+            return Optional.empty();
```
dominik-cnx, line 49: Is this a bit dangerous? I mean, the security won't kick if there are configuration issues and we don't easily see it? We maybe have the option to let the requests fail in such cases.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfigurationParser.java
```diff
@@ -0,0 +1,113 @@
...
+        Map<String, OAuthConfiguration> oAuthConfigurationsMap = new HashMap<>();
+
+        for (String authConfigId : authConfigs.fieldNames()) {
+            JsonObject authConfig = authConfigs.getJsonObject(authConfigId);
+
+            String flowTypeStr = authConfig.getString(FLOWTYPE_PROPERTY);
+            OAuth2FlowType flowType = flowTypeFromStr(flowTypeStr);
+
+            if(flowType == null) {
+                log.warn("No valid OAuth2FlowType configured for auth configuration '{}'. " +
```
dominik-cnx, line 67: dito

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
```diff
@@ -221,6 +131,13 @@ public Router(Vertx vertx,
         this.httpClientFactory = httpClientFactory;
         this.doneHandlers = doneHandlers;
         this.routeMultiplier = routeMultiplier;
+        this.oAuthProvider = oAuthProvider;
+
+        if(oAuthProvider != null) {
```
dominik-cnx, line 136: code format

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/auth/AuthHeader.java
```diff
@@ -0,0 +1,36 @@
+package org.swisspush.gateleen.routing.auth;
+
+public class AuthHeader {
+    private final String key = "Authorization";
+    private final String value;
```
dominik-cnx, line 5: probably not relevant for us here, but basically HTTP headers allow multiple values (per key)

dominik-cnx:
I like the implementation, thank you. The build currently doesn't pass, I haven't investigated why.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
mcweba: I changed the status code to _401 UNAUTHORIZED_

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
mcweba: I the _RuleFactory_ class I added
```java
if (basicAuth != null && oAuthId != null) {
	throw new ValidationException("Either 'basicAuth' or 'oAuthId' can be given, not both");
}
```
on purpose. Normally we do already in the JSON schema enforce that only one of _basicAuth_ or _oAuthId_ can be provided. With this additional check I wanted to make it even more clear, that only one at a time is allowed. Also the unit test is more readable :-)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfiguration.java
mcweba: Going to introduce a dedicated class for this id called _OAuthId_ instead of just a _String_

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfigurationParser.java
mcweba: When there is an issue with the router configuration resource, we won't have any configurations to be used. When a routing rule with a configured _oAuthId_ is used, the request will fail because no configuration can be found. Also when an endpoint that is protected with OAuth is called without authorization, the request will be rejected anyways.

Therefore, I think the current implementation is OK. What do you think?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/RouterConfigurationParser.java
mcweba: The same thing here because only the invalid configuration will not be used and rules using this configuration will also receive failing requests.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/auth/AuthHeader.java
mcweba: You are right. I also think it is not relevant here. I don't see a case where we would have multiple tokens.

PR #513, opened 2023-05-24
Removed references to post internal resources

dominik-cnx:


gateleen-hook-js/pom.xml
```diff
@@ -17,9 +17,7 @@
     </dependencies>
 
     <properties>
-        <nodeDownloadRoot>http://artifactory.pnet.ch/artifactory/list/ext-binaries-local/node/</nodeDownloadRoot>
```
dominik-cnx, line 20: This means that all devs now have to use the public repos

mcweba (approved):

PR #557, opened 2024-02-06
Feature/550 queue splitter

gcastaldi:


gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            log.info("Queue splitter configuration resource {} was updated. Going to initialize with new configuration", resourceUri);
+            System.out.println("Queue splitter configuration resource changed");
+            initializeQueueSplitterConfiguration(resource);
+        }
+    }
+
+    @Override
+    public void resourceRemoved(String resourceUri) {
+        if (configResourceUri() != null && configResourceUri().equals(resourceUri)) {
+            log.info("Queue splitter configuration resource {} was removed. Going to close all kafka producers", resourceUri);
```
mcweba, line 99: Change wrong log statement (kafka producers).

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
```diff
@@ -0,0 +1,105 @@
...
+            System.out.println("Queue splitter configuration resource removed");
```
mcweba, line 100: Remove System.out.println

gateleen-queue/src/main/resources/gateleen_queue_splitter_configuration_schema
```diff
@@ -0,0 +1,32 @@
+{
```
mcweba, line 1: Do not allow other properties to be added. Use `additionalProperties=false`

gateleen-queue/src/main/resources/gateleen_queue_splitter_configuration_schema
```diff
@@ -0,0 +1,32 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
```
mcweba, line 2: Add a `description` property to all the properties

gateleen-queue/src/main/resources/gateleen_queue_splitter_configuration_schema
```diff
@@ -0,0 +1,32 @@
...
+  "type": "object",
+  "additionalProperties": {
+    "$ref": "#/definitions/QueueSplitter"
+  },
+  "definitions": {
```
mcweba, line 7: Try to add more schema constraints like `oneOf` to clearify that only postfixFromStatic **or** (postfixFromHeader or postfixFromUrl) can be used. Checkout the schema defined for `gateleen-delegates` where this is used.

gateleen-queue/src/test/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImplTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        // Given
+        Async async = context.async();
+        context.assertFalse(queueSplitter.isInitialized());
+
+        // When
+        queueSplitter.initialize().onComplete(event -> {
+
+            // Then
+            context.assertFalse(queueSplitter.isInitialized());
```
mcweba, line 61: Shouldn't the `queueSplitter` be initialized now?

gateleen-runconfig/src/main/java/org/swisspush/gateleen/runconfig/RunConfig.java
```diff
@@ -597,11 +608,25 @@ private void handleRequest(final RoutingContext ctx) {
...
                 } else {
                     if (QueuingHandler.isQueued(request)) {
                         setISO8601Timestamps(request);
-                        request.bodyHandler(new QueuingHandler(vertx, redisProvider, request, monitoringHandler));
+                        request.bodyHandler(new QueuingHandler(
+                                vertx,
+                                redisProvider,
+                                request,
+                                monitoringHandler,
+                                queueSplitter != null ? queueSplitter : new NoOpQueueSplitter()));
```
mcweba, line 629: You already do this logic in the QueuingHandler constructor. I would suggest to just use the `queueSplitter` even if null and do this check in the Constructor. When `null` is passed you could create a `NoOpQueueSplitter`

mcweba (changes requested):

gateleen-queue/src/test/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImplTest.java
gcastaldi: If the configuration is missing or invalid the QueueSplitterIml is not initialized

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterConfigurationParser.java
```diff
@@ -60,9 +61,10 @@ static List<QueueSplitterConfiguration> parse(Buffer configurationResourceBuffer
                             null
                     ));
                 } else {
-                    String postfixFromHeader = queueConfig.getString(POSTFIX_FROM_HEADER_KEY);
-                    String postfixFromUrl = queueConfig.getString(POSTFIX_FROM_URL_KEY);
-                    if (postfixFromHeader != null || postfixFromUrl != null) {
+                    JsonObject postfixFromRequest = queueConfig.getJsonObject(POSTFIX_FROM_REQUEST_KEY);
+                    String postfixFromHeader = postfixFromRequest != null ? postfixFromRequest.getString(POSTFIX_FROM_HEADER_KEY) : null;
+                    String postfixFromUrl = postfixFromRequest != null ? postfixFromRequest.getString(POSTFIX_FROM_URL_KEY) : null;
+                    if (postfixFromRequest != null && (postfixFromHeader != null || postfixFromUrl != null)) {
```
mcweba, line 67: `postfixFromRequest != null` can be omitted since you already check `postfixFromHeader` and `postfixFromUrl`

gateleen-queue/src/main/resources/gateleen_queue_splitter_configuration_schema
```diff
@@ -1,5 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "A list of queue splitter configurations",
```
mcweba, line 3: queue splitter configurations are not in a list ;-)

gateleen-queue/src/main/resources/gateleen_queue_splitter_configuration_schema
mcweba, line 1: You could also add unit tests for the schema validation. See `KafkaTopicConfigurationSchemaValidationTest` as example

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
```diff
@@ -0,0 +1,103 @@
...
+                promise.complete();
+            } else {
+                log.warn("No queue splitter configuration resource with uri '{}' found. Unable to setup splitter configuration correctly", configResourceUri());
+                promise.complete();
+            }
+        }));
+        return promise.future();
+    }
+
+    public boolean isInitialized() {
```
mcweba, line 61: What do you need
```java
public boolean isInitialized() {
    return initialized;
}
```
for? Only for tests? What happens if you remove the config resource and `initialized=false` is set?

gcastaldi: Yes, the method is used only in QueueSplitterImpl tests to verify the current configuration is valid. Same purpose as KafkaHandler.isInitialized() also used only in KafkaHandler tests.

mcweba (approved):

PR #562, opened 2024-03-08
[SDCISA-13746] Increase resilience on resource exhaustion

hiddenalpha:
The issue currently ([2024-03-08](https://xkcd.com/1179/)) can be observed here:
> oc -n isa-houston-test exec -ti "$(oc -n isa-houston-test get pods|egrep ston-[0-9]|egrep -v deplo|cut -f1 -d' ')" -- sh -c 'cd /usr/local/vertx/logs && tail -n42 -F houston.log | grep ConnectionPoolTooBusyException'

### DIFF
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -264,7 +264,7 @@ public class Forwarder extends AbstractForwarder {
                     long postponeMs = 1000;
                     long remainingTimeMs = timeoutMs - postponeMs;
                     if (remainingTimeMs > 0 && ex instanceof ConnectionPoolTooBusyException) {
-                        log.debug("EAGAIN: No connection avail now for {}", targetUri, ex);
+                        log.debug("No connection avail now for {}", targetUri, ex);
                         vertx.setTimer(postponeMs, nonsense -> {
                             retryPerformRequest(req, bodyData, targetUri, log, profileHeaderMap, authHeader,
                                 afterHandler, loggingHandler, uniqueId, remainingTimeMs, startTime);
--- a/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
+++ b/gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
@@ -245,13 +245,13 @@ public class Forwarder extends AbstractForwarder {
         long timeoutMs = (timeoutRaw != null) ? parseLong(timeoutRaw) : rule.getTimeout();
         final long startTime = monitoringHandler.startRequestMetricTracking(rule.getMetricName(), req.uri());
 
-        retryPerformRequest(req, bodyData, targetUri, log, profileHeaderMap, authHeader.orElse(null),
+        retryPerformRequest(req, bodyData, targetUri, log, profileHeaderMap, authHeader,
             afterHandler, loggingHandler, uniqueId, timeoutMs, startTime);
     }
 
     private void retryPerformRequest(
         HttpServerRequest req, Buffer bodyData, String targetUri, Logger log, Map<String, String> profileHeaderMap,
-        AuthHeader authHeader, @Nullable Handler<Void> afterHandler, LoggingHandler loggingHandler,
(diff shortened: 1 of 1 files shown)


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -239,31 +241,49 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
...
             public void handle(AsyncResult<HttpClientRequest> event) {
                 req.resume();
 
                 if (event.failed()) {
-                    log.warn("Problem to request {}: {}", targetUri, event.cause());
+                    Throwable ex = event.cause();
+                    long postponeMs = 1000;
+                    long remainingTimeMs = timeoutMs - postponeMs;
+                    if (remainingTimeMs > 0 && ex instanceof ConnectionPoolTooBusyException) {
+                        log.debug("EAGAIN: No connection avail now for {}", targetUri, ex);
```
Kusig, line 267: Why EAGAIN, is this a typo?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -239,31 +241,49 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
...
             @Override
             public void handle(AsyncResult<HttpClientRequest> event) {
                 req.resume();
 
                 if (event.failed()) {
-                    log.warn("Problem to request {}: {}", targetUri, event.cause());
+                    Throwable ex = event.cause();
+                    long postponeMs = 1000;
+                    long remainingTimeMs = timeoutMs - postponeMs;
+                    if (remainingTimeMs > 0 && ex instanceof ConnectionPoolTooBusyException) {
```
Kusig, line 266: What about the sizing of the ConnectionPool itself, must it be tuned accordingly to reduce the probability of this occurence?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
Kusig, line 267: Would this be worth of any metric in order to monitor the ConnectionPool Failures over time?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
hiddenalpha: Feel free to open a PR for it :)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
hiddenalpha: If so, we should measure how loaded the pool is. Not how often we end up in this code branch.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -278,7 +298,7 @@ public void handle(AsyncResult<HttpClientRequest> event) {
                 }
                 setProfileHeaders(log, profileHeaderMap, cReq);
 
-                authHeader.ifPresent(authHeaderValue -> cReq.headers().set(authHeaderValue.key(), authHeaderValue.value()));
+                if (authHeader != null) cReq.headers().set(authHeader.key(), authHeader.value());
```
mcweba, line 301: Why going from _Optional<AuthHeader>_ to _AuthHeader_ where you have to use _null_ checks again. I don't see how this makes the code better or more readable

hiddenalpha: My opinion clearly differs in this regard. I just changed it back.

(HINT: Feel free to ignore my opinion that follows here)
<opinion>

Optional is nothing but adding unnecessary bloat in this case (*Just comparing both variants side-by-side should be enough to show why*). [I'm not alone](https://homes.cs.washington.edu/~mernst/advice/nothing-is-better-than-optional.html) with this opinion BTW. Yet another counter argument comes from IDEA itslef:

image

</opinion>

PR #573, opened 2024-05-02
Handle "Payload Too Large" responses from vertx-rest-storage

mcweba:


dominik-cnx (approved):

gateleen-expansion/src/main/java/org/swisspush/gateleen/expansion/ExpansionHandler.java
```diff
@@ -544,8 +545,11 @@ private void makeStorageExpandRequest(final String targetUri, final List subReso
                     return;
                 }
                 cRes.bodyHandler(data -> {
-                    if (StatusCode.INTERNAL_SERVER_ERROR.getStatusCode() == cRes.statusCode()) {
-                        String fullResponseBody = data.toString();
+                    String fullResponseBody = data.toString();
```
hiddenalpha, line 548: **WARN**: Moving this outside the condition causes it to be executed for **ALL** cases. So we'll likely setup lots of pressure towards GC. I think it would be better to duplicate this line into both of the first two if cases instead.

mcweba: Good point, I will move it

PR #577, opened 2024-05-21
Reduce risk of Logging Resource exhaustion

hiddenalpha:


mcweba (changes requested):
Changes all look good. Before merging this to develop I would suggest to replace all occurrences of logging `new Exception(...)` with the new factory method

Issue #49, opened 2016-07-01
QueueBrowser: QueueItems with index higher than 99 cannot be loaded


mcweba:
In the QueueBrowser class is a mapping for GET requests to load a single QueueItem by index.

> GET /playground/queues/my_queue/{index}

This mapping is implemented as follows:

``` java
// Get item
router.getWithRegex(prefix + "/queues/([^/]+)/[0-9]+").handler(ctx -> {
    final String queue = lastPart(ctx.request().path().substring(0, ctx.request().path().length() - 2), "/");
    final int index = Integer.parseInt(lastPart(ctx.request().path(), "/"));
    eb.send(redisquesAddress, buildGetQueueItemOperation(queue, index), new Handler<AsyncResult<Message<JsonObject>>>() {
        @Override
        public void handle(AsyncResult<Message<JsonObject>> reply) {
            JsonObject replyBody = reply.result().body();
            if (OK.equals(replyBody.getString(STATUS))) {
                ctx.response().putHeader(CONTENT_TYPE, APPLICATION_JSON);
                ctx.response().end(decode(reply.result().body().getString(VALUE)));
            } else {
                ctx.response().setStatusCode(StatusCode.NOT_FOUND.getStatusCode());
                ctx.response().setStatusMessage(StatusCode.NOT_FOUND.getStatusMessage());
                ctx.response().end("Not Found");
            }
        }
    });
});
```

The following line is wrong, since it only works for indices from 0 to 99

``` java
final String queue = lastPart(ctx.request().path().substring(0, ctx.request().path().length() - 2), "/");
```

Making a request like this

> GET /playground/queues/my_queue/500

would result in a queuename with value _5_ instead of _my_queue_.

mcweba:
Fixed in [v2.1.8](https://github.com/swisspost/gateleen/releases/tag/v2.1.8) by removing the deprecated QueueBrowser class

PR #581, opened 2024-05-29
#49 Fixed bug in QueueBrowser by deleting it. Is deprecated since 2021

mcweba:


hiddenalpha:
I don't know this class. I guess its ok. Maybe @ljucam knows more?

hiddenalpha (approved):

PR #599, opened 2024-06-18
write more detail log if forwarding to storage fails

dominik-cnx:


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Router.java
```diff
@@ -85,6 +86,8 @@ public class Router implements Refreshable, LoggableResource, ConfigurationResou
     private OAuthStrategy oAuthStrategy = null;
     private BasicAuthStrategy basicAuthStrategy;
 
+    private static GateleenExceptionFactory gateleenExceptionFactory = GateleenExceptionFactory.newGateleenThriftyExceptionFactory();
```
hiddenalpha, line 89: I strongly advise to use constructor injection. We Must be able to hand in our custom implementation from application side.

Using it hardcoded, defeats the whole purpose of it.

mcweba: I agree on this comment

mcweba (approved):

PR #613, opened 2024-10-04
Issue311 improve hook searching

steniobhz:
### Documentation for the Hook Search API (Listeners and Routes)

Available Endpoints
1. Search Listeners

- Endpoint: /registrations/listeners
- HTTP Method: GET
- Description: Searches for all Listeners whose destination field contains the provided query string in the q parameter.

2. Search Routes

- Endpoint: /registrations/routes
- HTTP Method: GET
- Description: Searches for all Routes whose destination field contains the provided query string in the q parameter.

Query Parameter

Both endpoints support the q query parameter, which is used to filter the search results.

- q: A string to search for within the destination field of Listeners or Routes.

Example Request:
GET /registrations/listeners?q=myDestination
GET /registrations/routes?q=myDestination

Response Structure
Example Response - Searching for Listeners:

```
{
    "listeners": [
        "listener1",
        "listener2",
        "listener3"
    ]
}

```
Example Response - Searching for Routes:

```
{
    "routes": [
        "route1",
        "route2",
        "route3"
    ]
}
```

Key Features

1. Separate Endpoints for Listeners and Routes: The API provides distinct endpoints for Listeners and Routes searches, ensuring optimized search results based on specific hook types.

2. String-Based Search: The API looks for matches in the destination field of both Listeners and Routes. It searches for any substring match rather than an exact match, making the search more flexible.

3. Optimized Search Logic: The API uses efficient search mechanisms internally to quickly find results across registered listeners or routes, reducing response time even with large datasets.

Example Workflow

1. A client wants to find all Listeners with a destination that contains the string "myDestination". They make the following request:
GET /registrations/listeners?q=myDestination

2. The API returns a JSON object containing the identifiers of matching Listeners:
```
{
    "listeners": [
        "listener1",
        "listener2"
    ]
}
```

### DIFF
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -248,7 +248,7 @@ Gateleen allows searching for listeners and routes using the query parameter `q`
 Search for listeners based on a query parameter like this:
 
 ```
-GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners?q=test
+GET http://myserver:7012/playground/server/hooks/v1/registrations/listeners?q=test
 ```
 
 The response will contain the matching listeners. If no match is found, an empty list is returned:
@@ -273,7 +273,7 @@ The response will contain the matching listeners. If no match is found, an empty
 Similarly, you can search for routes using a query parameter:
 
 ```
-GET http://myserver:7012/gateleen/server/routetest/_hooks/routes?q=test
+GET http://myserver:7012/playground/server/hooks/v1/registrations/routes/?q=test
 ```
 
 The response contains the matching routes, or an empty list if no match is found.
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -240,10 +240,43 @@ hookHandler.enableResourceLogging(true);
 ```
 
 
+## Query-Based Listener and Route Search
 
+Gateleen allows searching for listeners and routes using the query parameter `q`. This simplifies filtering the registered hooks based on query parameters.
 
+### Listener Search with `q`
+Search for listeners based on a query parameter like this:
 
+```
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners/listener/1?q=testQuery
+```
 
+The response will contain the matching listeners. If no match is found, an empty list is returned:
 
+**Example response with matches:**
+```json
+{
+  "listeners": [
+    {
+      "destination": "/path/to/destination",
+      "methods": ["GET", "POST"]
+    }
+  ]
+}
+```
 
+**Example response with no matches:**
+```json
+{
+  "listeners": []
+}
+```
 
+### Route Search with `q`
+Similarly, you can search for routes using a query parameter:
+
+```
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/routes?q=testRoute
+```
+
+The response contains the matching routes, or an empty list if no match is found.
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -244,11 +244,13 @@ hookHandler.enableResourceLogging(true);
 
 Gateleen allows searching for listeners and routes using the query parameter `q`. This simplifies filtering the registered hooks based on query parameters.
 
+The search will be based on the value registered of the destination property
+
 ### Listener Search with `q`
 Search for listeners based on a query parameter like this:
 
 ```
-GET http://myserver:7012/playground/server/hooks/v1/registrations/listeners?q=test
+GET http://myserver:7012/playground/server/hooks/v1/registrations/listeners?q=mylistener
 ```
 
 The response will contain the matching listeners. If no match is found, an empty list is returned:
@@ -257,7 +259,7 @@ The response will contain the matching listeners. If no match is found, an empty
 ```json
 {
   "listeners": [
-    "first+playground+server+test+nemo+origin+b"
+    "first+playground+server+test+nemo+origin+mylistener"
   ]
 }
 ```
@@ -273,7 +275,23 @@ The response will contain the matching listeners. If no match is found, an empty
 Similarly, you can search for routes using a query parameter:
 
 ```
-GET http://myserver:7012/playground/server/hooks/v1/registrations/routes/?q=test
+GET http://myserver:7012/playground/server/hooks/v1/registrations/routes/?q=myroute
 ```
 
 The response contains the matching routes, or an empty list if no match is found.
+
+**Example response with matches:**
+```json
+{
+  "routes": [
+    "first+playground+server+test+nemo+origin+myroute"
+  ]
+}
+
+```
+**Example response with no matches:**
+```json
+{
+  "routes": []
+}
+```
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -257,10 +257,7 @@ The response will contain the matching listeners. If no match is found, an empty
 ```json
 {
   "listeners": [
-    {
-      "destination": "/path/to/destination",
-      "methods": ["GET", "POST"]
-    }
+    "first+playground+server+test+nemo+origin+b"
   ]
 }
 ```
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -248,7 +248,7 @@ Gateleen allows searching for listeners and routes using the query parameter `q`
 Search for listeners based on a query parameter like this:
 
 ```
-GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners/listener/1?q=testQuery
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners?q=testQuery
 ```
 
 The response will contain the matching listeners. If no match is found, an empty list is returned:
--- a/gateleen-hook/README_hook.md
+++ b/gateleen-hook/README_hook.md
@@ -248,7 +248,7 @@ Gateleen allows searching for listeners and routes using the query parameter `q`
 Search for listeners based on a query parameter like this:
 
 ```
-GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners?q=testQuery
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners?q=test
 ```
 
 The response will contain the matching listeners. If no match is found, an empty list is returned:
@@ -273,7 +273,7 @@ The response will contain the matching listeners. If no match is found, an empty
 Similarly, you can search for routes using a query parameter:
 
 ```
-GET http://myserver:7012/gateleen/server/listenertest/_hooks/routes?q=testRoute
+GET http://myserver:7012/gateleen/server/routetest/_hooks/routes?q=test
 ```
 
 The response contains the matching routes, or an empty list if no match is found.
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -569,6 +569,14 @@ public class HookHandler implements LoggableResource {
             }
         }
 
+        HttpServerResponse response = ctx.response();
+        String queryParam = request.getParam("q");
+
+        if ((queryParam != null) && !queryParam.isEmpty()) {
+            this.handleHookSearch(queryParam,response);
+            return true;
+        }
+
         /*
          * 2) Check if we have to queue a request for listeners
          */
@@ -591,6 +599,39 @@ public class HookHandler implements LoggableResource {
             return true;
         }
     }
+
+    /**
+     * Handles hook search requests based on the 'destination' property.
+     * Searches in both routes and listeners.
+     *
+     * @param queryParam the RoutingContext of the request
+     */
+    public void handleHookSearch(String queryParam,HttpServerResponse response) {
+        JsonObject result = new JsonObject();
+        JsonArray matchingRoutes = new JsonArray();
+        JsonArray matchingListeners = new JsonArray();
+
+        // Search routes by destination
+        routeRepository.getRoutes().forEach((routeKey, route) -> {
+            if (route.getHook().getDestination().contains(queryParam)) {
+                matchingRoutes.add(routeKey);
+            }
+        });
+
+        // Search listeners by destination
+        listenerRepository.getListeners().forEach(listener -> {
+            if (listener.getHook().getDestination().contains(queryParam)) {
+                matchingListeners.add(listener.getListenerId());
+            }
+        });
+
+        // Build and send the response
+        result.put("routes", matchingRoutes);
+        result.put("listeners", matchingListeners);
+
+        response.putHeader("content-type", "application/json").end(result.encode());
+    }
+
 
     /**
      * Create a listing of routes in the given parent. This happens
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -577,11 +577,11 @@ public class HookHandler implements LoggableResource {
 
         if (requestMethod == GET && !request.params().isEmpty()) {
             String queryParam = request.getParam("q");
-            if (requestUri.contains(normalizedListenerBase) && checkValidParams(request,queryParam)) {
-                handleListenerSearch(queryParam, request.response());
+            if (requestUri.contains(normalizedListenerBase) ) {
+                handleListenerSearch(queryParam, request);
                 return true;
-            } else if (requestUri.contains(normalizedRouteBase) && checkValidParams(request,queryParam)) {
-                handleRouteSearch(queryParam, request.response());
+            } else if (requestUri.contains(normalizedRouteBase) ) {
+                handleRouteSearch(queryParam, request);
                 return true;
             }
         }
@@ -609,32 +609,23 @@ public class HookHandler implements LoggableResource {
         }
     }
 
-    private boolean checkValidParams (HttpServerRequest request, String queryParam){
-        if (request.params().size() > 1 || (queryParam == null) || (queryParam.isEmpty())) {
-            request.response().setStatusCode(StatusCode.BAD_REQUEST.getStatusCode())
-                    .end("Bad Request: Only the 'q' parameter is allowed and can't be empty or null");
-            return true;
-        }
-        return false;
-    }
-
-    private void handleListenerSearch(String queryParam, HttpServerResponse response) {
+    private void handleListenerSearch(String queryParam, HttpServerRequest request) {
         handleSearch(
                 listenerRepository.getListeners().stream().collect(Collectors.toMap(Listener::getListenerId, listener -> listener)),
                 listener -> listener.getHook().getDestination(),
                 queryParam,
                 LISTENERS_KEY,
-                response
+                request
         );
     }
 
-    private void handleRouteSearch(String queryParam, HttpServerResponse response) {
+    private void handleRouteSearch(String queryParam, HttpServerRequest request) {
         handleSearch(
                 routeRepository.getRoutes().entrySet().stream().collect(Collectors.toMap(entry -> entry.getValue().getHookIdentify(), Map.Entry::getValue)),
                 route -> route.getHook().getDestination(),
                 queryParam,
                 ROUTES_KEY,
-                response
+                request
         );
     }
 
@@ -647,9 +638,14 @@ public class HookHandler implements LoggableResource {
      * @param getDestination Function to extract destinations.
      * @param queryParam The query string to match.
      * @param resultKey The key for the result in the response.
-     * @param response The HTTP response to return the results. Must not be null.
+     * @param request The HTTP request to make a specific validations and return the results.
      */
-    private <T> void handleSearch(Map<String, T> repository, Function<T, String> getDestination, String queryParam, String resultKey, HttpServerResponse response) {
+    private <T> void handleSearch(Map<String, T> repository, Function<T, String> getDestination, String queryParam, String resultKey, HttpServerRequest request) {
+        if (request.params().size() > 1 || (queryParam == null) || (queryParam.isEmpty())) {
+            request.response().setStatusCode(StatusCode.BAD_REQUEST.getStatusCode())
+                    .end("Bad Request: Only the 'q' parameter is allowed and can't be empty or null");
+            return ;
+        }
 
         JsonArray matchingResults = new JsonArray();
         repository.forEach((key, value) -> {
@@ -664,9 +660,10 @@ public class HookHandler implements LoggableResource {
 
         String encodedResult = result.encode();
 
-        response.putHeader(CONTENT_TYPE_HEADER, CONTENT_TYPE_JSON);
-        response.end(encodedResult);
+        request.response().putHeader(CONTENT_TYPE_HEADER, CONTENT_TYPE_JSON);
+        request.response().end(encodedResult);
     }
+
 
 
     /**
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -583,10 +583,10 @@ public class HookHandler implements LoggableResource {
             String queryParam = request.getParam("q");
             // 2. Check if the URI is for listeners or routes and has a query parameter
             if (queryParam != null && !queryParam.isEmpty()) {
-                if (uri.contains(HOOK_LISTENER_STORAGE_PATH)) {
+                if (uri.contains(LISTENERS_KEY)) {
                     handleListenerSearch(queryParam, request.response());
                     return true;
-                } else if (uri.contains(HOOK_ROUTE_STORAGE_PATH)) {
+                } else if (uri.contains(ROUTES_KEY)) {
                     handleRouteSearch(queryParam, request.response());
                     return true;
                 }
@@ -649,11 +649,15 @@ public class HookHandler implements LoggableResource {
         JsonObject result = new JsonObject();
         result.put(resultKey, matchingResults);
 
-        // Set headers safely before writing the response
-        response.putHeader(CONTENT_TYPE_HEADER, CONTENT_TYPE_JSON);
-        response.write(result.encode());
-        response.end();
+        String encodedResult = result.encode(); // Convert the result to a string
 
+        // Set Content-Length header before sending the response
+        response.putHeader(CONTENT_TYPE_HEADER, CONTENT_TYPE_JSON);
+        response.putHeader("Content-Length", String.valueOf(encodedResult.length())); // Set content length
+
+        // Write and end the response
+        response.write(encodedResult);
+        response.end();
     }
 
     /**
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -90,7 +90,6 @@ public class HookHandler implements LoggableResource {
     public static final String HOOKS_LISTENERS_URI_PART = "/_hooks/listeners/";
     public static final String LISTENER_QUEUE_PREFIX = "listener-hook";
     private static final String X_QUEUE = "x-queue";
-    private static final String X_EXPIRE_AFTER = "X-Expire-After";
     private static final String LISTENER_HOOK_TARGET_PATH = "listeners/";
 
     public static final String HOOKS_ROUTE_URI_PART = "/_hooks/route";
@@ -127,7 +126,6 @@ public class HookHandler implements LoggableResource {
     private static final String CONTENT_TYPE_JSON = "application/json";
     private static final String LISTENERS_KEY = "listeners";
     private static final String ROUTES_KEY = "routes";
-    private static final String DESTINATION_KEY = "destination";
     private static final String CONTENT_TYPE_HEADER = "content-type";
 
 
@@ -549,13 +547,12 @@ public class HookHandler implements LoggableResource {
     public boolean handle(final RoutingContext ctx) {
         HttpServerRequest request = ctx.request();
         boolean consumed = false;
-
+        var requestUri = request.uri();
         /*
          * 1) Un- / Register Listener / Routes
          */
         var requestMethod = request.method();
         if (requestMethod == PUT) {
-            var requestUri = request.uri();
             if (requestUri.contains(HOOKS_LISTENERS_URI_PART)) {
                 handleListenerRegistration(request);
                 return true;
@@ -566,7 +563,6 @@ public class HookHandler implements LoggableResource {
             }
         }
         if (requestMethod == DELETE) {
-            var requestUri = request.uri();
             if (requestUri.contains(HOOKS_LISTENERS_URI_PART)) {
                 handleListenerUnregistration(request);
                 return true;
@@ -579,15 +575,20 @@ public class HookHandler implements LoggableResource {
 
         // 1. Check if the request method is GET
         if (request.method() == HttpMethod.GET) {
-            String uri = request.uri();
             String queryParam = request.getParam("q");
-            // 2. Check if the URI is for listeners or routes and has a query parameter
-            if (queryParam != null && !queryParam.isEmpty()) {
-                if (uri.contains(LISTENERS_KEY)) {
+            // If the 'q' parameter exists, proceed with search handling
+            if (queryParam != null) {
+                // Check if the URI corresponds to listeners or routes
+                if (requestUri.contains(LISTENERS_KEY)) {
                     handleListenerSearch(queryParam, request.response());
                     return true;
-                } else if (uri.contains(ROUTES_KEY)) {
+                } else if (requestUri.contains(ROUTES_KEY)) {
                     handleRouteSearch(queryParam, request.response());
+                    return true;
+                }
+            }else{
+                if (!request.params().isEmpty()) {
+                    request.response().setStatusCode(400).end("Bad Request: Only the 'q' parameter is allowed");
                     return true;
                 }
             }
@@ -649,8 +650,9 @@ public class HookHandler implements LoggableResource {
      * @param response The HTTP response to return the results. Must not be null.
      */
     private <T> void handleSearch(Map<String, T> repository, Function<T, String> getDestination, String queryParam, String resultKey, HttpServerResponse response) {
-        if (repository == null || getDestination == null) {
-            response.setStatusCode(400).end(); // Bad request for missing parameters
+
+        if (repository == null || getDestination == null || resultKey == null || queryParam.isEmpty()) {
+            response.setStatusCode(400).end("Bad Request: One or more required parameters are missing or null");
             return;
         }
 
@@ -658,7 +660,7 @@ public class HookHandler implements LoggableResource {
 
         repository.forEach((key, value) -> {
             String destination = getDestination.apply(value);
-            if (destination != null && destination.contains(queryParam != null ? queryParam : "")) {
+            if (destination != null && destination.contains(queryParam)) {
                 matchingResults.add(key);
             }
         });
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -581,7 +581,6 @@ public class HookHandler implements LoggableResource {
         if (request.method() == HttpMethod.GET) {
             String uri = request.uri();
             String queryParam = request.getParam("q");
-
             // 2. Check if the URI is for listeners or routes and has a query parameter
             if (queryParam != null && !queryParam.isEmpty()) {
                 if (uri.contains(HOOK_LISTENER_STORAGE_PATH)) {
@@ -591,9 +590,6 @@ public class HookHandler implements LoggableResource {
                     handleRouteSearch(queryParam, request.response());
                     return true;
                 }
-            }
-            else {
-                return false;
             }
         }
 
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -31,12 +31,7 @@ import org.swisspush.gateleen.core.http.HttpRequest;
 import org.swisspush.gateleen.core.logging.LoggableResource;
 import org.swisspush.gateleen.core.logging.RequestLogger;
 import org.swisspush.gateleen.core.storage.ResourceStorage;
-import org.swisspush.gateleen.core.util.CollectionContentComparator;
-import org.swisspush.gateleen.core.util.HttpHeaderUtil;
-import org.swisspush.gateleen.core.util.HttpRequestHeader;
-import org.swisspush.gateleen.core.util.HttpServerRequestUtil;
-import org.swisspush.gateleen.core.util.ResourcesUtils;
-import org.swisspush.gateleen.core.util.StatusCode;
+import org.swisspush.gateleen.core.util.*;
 import org.swisspush.gateleen.hook.queueingstrategy.DefaultQueueingStrategy;
 import org.swisspush.gateleen.hook.queueingstrategy.DiscardPayloadQueueingStrategy;
 import org.swisspush.gateleen.hook.queueingstrategy.QueueingStrategy;
@@ -576,12 +571,11 @@ public class HookHandler implements LoggableResource {
         }
 
         if (requestMethod == GET && !request.params().isEmpty()) {
-            String queryParam = request.getParam("q");
             if (requestUri.contains(normalizedListenerBase) ) {
-                handleListenerSearch(queryParam, request);
+                handleListenerSearch(request);
                 return true;
             } else if (requestUri.contains(normalizedRouteBase) ) {
-                handleRouteSearch(queryParam, request);
+                handleRouteSearch(request);
                 return true;
             }
         }
@@ -609,21 +603,19 @@ public class HookHandler implements LoggableResource {
         }
     }
 
-    private void handleListenerSearch(String queryParam, HttpServerRequest request) {
+    private void handleListenerSearch(HttpServerRequest request) {
         handleSearch(
                 listenerRepository.getListeners().stream().collect(Collectors.toMap(Listener::getListenerId, listener -> listener)),
                 listener -> listener.getHook().getDestination(),
-                queryParam,
                 LISTENERS_KEY,
                 request
         );
     }
 
-    private void handleRouteSearch(String queryParam, HttpServerRequest request) {
+    private void handleRouteSearch(HttpServerRequest request) {
         handleSearch(
                 routeRepository.getRoutes().entrySet().stream().collect(Collectors.toMap(entry -> entry.getValue().getHookIdentify(), Map.Entry::getValue)),
                 route -> route.getHook().getDestination(),
-                queryParam,
                 ROUTES_KEY,
                 request
         );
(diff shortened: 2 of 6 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -569,6 +569,18 @@ public boolean handle(final RoutingContext ctx) {
...
 
+        HttpServerResponse response = ctx.response();
+        String queryParam = null;
+
+        if (request.params() != null) {
+            queryParam = request.getParam("q");
+        }
+
+        if (queryParam != null && !queryParam.isEmpty()) {
+            this.handleHookSearch(queryParam, response);
```
mcweba, line 580: The `handleHookSearch` should be executed for `GET` requests to `/registrations/listeners` or `/registrations/routes` only. A request to the hookUrl will trigger the same code and there we don't want to search for hooks

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -592,6 +604,39 @@ public boolean handle(final RoutingContext ctx) {
         }
     }
 
+    /**
+     * Handles hook search requests based on the 'destination' property.
+     * Searches in both routes and listeners.
+     *
+     * @param queryParam the RoutingContext of the request
+     */
+    public void handleHookSearch(String queryParam,HttpServerResponse response) {
```
mcweba, line 613: Why do you search for routes and listeners? It is not very efficient when you make a `GET` to `/playground/server/hooks/v1/registrations/listeners?q=foobar` and then search for matching routes.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -592,6 +604,39 @@ public boolean handle(final RoutingContext ctx) {
...
+
+        // Search listeners by destination
+        listenerRepository.getListeners().forEach(listener -> {
+            if (listener.getHook().getDestination().contains(queryParam)) {
+                matchingListeners.add(listener.getListenerId());
+            }
+        });
+
+        // Build and send the response
+        result.put("routes", matchingRoutes);
```
mcweba, line 633: Please don't change the API. You have either a `routes` array or a `listeners` array, but never both. Be aware that such changes in the API can break a client using the API

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +644,144 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+        Mockito.when(listenerRepository.getListeners()).thenReturn(Collections.singletonList(mockListener));
+
+        // Create HookHandler instance
+        HookHandler hookHandler = new HookHandler(vertx, httpClient, storage, loggingResourceManager,
+                logAppenderRepository, monitoringHandler, "userProfilePath", HOOK_ROOT_URI, requestQueue,
+                false
+        );
+
+        // Use reflection to set private fields
+        setPrivateField(hookHandler, "listenerRepository", listenerRepository);
```
mcweba, line 680: Don't use reflection. You have mockable dependencies like `listenerRepository` and `routeRepository`. When there is no matching constructor in the `HookHandler` class, create one.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
mcweba, line 613: I think `handleHookSearch` should be private. There is no need to provide this method outside of `HookHandler` class

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +644,144 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+
+        HttpServerRequest request = Mockito.mock(HttpServerRequest.class);
+        Mockito.when(request.response()).thenReturn(response);
+        Mockito.when(request.getParam("q")).thenReturn(null);
+
+        // Call hookHandleSearch
+        hookHandler.handleHookSearch(null, response);
+
+        // Verify that nothing is returned
+        Mockito.verify(response, Mockito.never()).end(any(Buffer.class));
```
mcweba, line 782: What happens to the original request when `response.end()` is never called?

mcweba (changes requested):
As discussed on the phone, you should also add some integration tests in the `gateleen-test` module. Where is the documentation of this new API feature?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok, I will create a separate Endpoints for Listeners and Routes

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: I will recreate this test, and cover it.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: Ok.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok I will validate this.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: We have a timeout in place, but I'll be refining this approach.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: ok.

gateleen-hook/README_hook.md
```diff
@@ -240,10 +240,43 @@ hookHandler.enableResourceLogging(true);
...
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners/listener/1?q=testQuery
+```
+
+The response will contain the matching listeners. If no match is found, an empty list is returned:
+
+**Example response with matches:**
+```json
+{
+  "listeners": [
+    {
```
mcweba, line 260: The content of searching listeners and routes is not an array of objects (with destination, methods properties) but an array containing a list of names (strings). How did you even get this response?

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -123,6 +124,13 @@ public class HookHandler implements LoggableResource {
     public static final String LISTABLE = "listable";
     public static final String COLLECTION = "collection";
 
+    private static final String CONTENT_TYPE_JSON = "application/json";
+    private static final String LISTENERS_KEY = "listeners";
+    private static final String ROUTES_KEY = "routes";
+    private static final String DESTINATION_KEY = "destination";
```
mcweba, line 130: Your newly created `DESTINATION_KEY` static variable is not used

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -569,6 +577,22 @@ public boolean handle(final RoutingContext ctx) {
             }
         }
 
+        // 1. Check if the request method is GET
+        if (request.method() == HttpMethod.GET) {
+            String uri = request.uri();
```
mcweba, line 582: Extract `var requestUri = request.uri();` only once in the `handle` method. It is already done for `PUT`and `DELETE` requests. We should optimize this.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -569,6 +577,22 @@ public boolean handle(final RoutingContext ctx) {
...
+            String queryParam = request.getParam("q");
```
mcweba, line 583: Using any other query params than `q` in `GET` requests should result in a `400 Bad Request` and not just return all listeners/hooks. Also create tests to verify this.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -569,6 +577,22 @@ public boolean handle(final RoutingContext ctx) {
...
+            // 2. Check if the URI is for listeners or routes and has a query parameter
+            if (queryParam != null && !queryParam.isEmpty()) {
+                if (uri.contains(LISTENERS_KEY)) {
```
mcweba, line 586: Don't look for `listeners` only. This will also match other requests not related to hooks. E.g.

```
/nemo/some/url/listeners/foobar
```
The same applies for `routes`. You always have a configurable prefix like `playground/server/hooks/v1/registrations` which should be used for this.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -592,6 +616,65 @@ public boolean handle(final RoutingContext ctx) {
...
+     *
+     * @param repository The items to search.
+     * @param getDestination Function to extract destinations.
+     * @param queryParam The query string to match.
+     * @param resultKey The key for the result in the response.
+     * @param response The HTTP response to return the results. Must not be null.
+     */
+    private <T> void handleSearch(Map<String, T> repository, Function<T, String> getDestination, String queryParam, String resultKey, HttpServerResponse response) {
+        if (repository == null || getDestination == null) {
+            response.setStatusCode(400).end(); // Bad request for missing parameters
```
mcweba, line 653: Use `StatusCode.BAD_REQUEST.getStatusCode()` instead of 400

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -592,6 +616,65 @@ public boolean handle(final RoutingContext ctx) {
...
+            }
+        });
+
+        JsonObject result = new JsonObject();
+        result.put(resultKey, matchingResults);
+
+        String encodedResult = result.encode();
+
+        response.putHeader(CONTENT_TYPE_HEADER, CONTENT_TYPE_JSON);
+        response.putHeader("Content-Length", String.valueOf(encodedResult.length()));
```
mcweba, line 672: The two lines are not needed:
```
response.putHeader("Content-Length", String.valueOf(encodedResult.length()));
response.write(encodedResult);
```
use `response.end(encodedResult)` directly.

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+
+        // Mock RoutingContext
+        RoutingContext routingContext = mock(RoutingContext.class);
+        when(routingContext.request()).thenReturn(request);
+
+        // Act
+        boolean result = hookHandler.handle(routingContext); // Calls the public `handle` method
+
+        // Assert
+        assertTrue(result); // Ensure the handler returns true for a valid listener search
```
mcweba, line 669: Your test does not verify that the search was actually successfull. It only tests whether the request was handled by the `HookHandler`. You should verify that the response is something like:
```json

{
    "listeners": []
}
```

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
         vertx.eventBus().request("gateleen.hook-route-remove", "pathToRouterResource");
     }
 
+    @Test
+    public void testHandleListenerSearch_Success() {
+        // Arrange
+        String queryParam = "validQueryParam";
+        String uri = "registrations/listeners/?q=" + queryParam;
```
mcweba, line 650: You have a duplicate queryParam here because of `when(request.getParam("q")).thenReturn(queryParam);`

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+        // Act
+        boolean result = hookHandler.handle(routingContext); // Calls the public `handle` method
+
+        // Assert
+        assertTrue(result); // Ensure the handler returns true for a valid listener search
+        verify(response, times(1)).end(); // Ensure `response.end()` was called
+    }
+    @Test
+    public void testHandleListenerSearch_MissingQueryParam() {
+        String uri = "registrations/listeners/?q=";
```
mcweba, line 674: Duplicate query param

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+        Async async = context.async();
+
+        // Act: Call the hookHandler.handle method
+        boolean handled = hookHandler.handle(routingContext);
+
+        // Assert that it was handled
+        context.assertTrue(handled);
+
+        // Verify that the response ended correctly (simulating a successful response)
+        Mockito.verify(response, Mockito.times(1)).end();
```
mcweba, line 731: Your test does not find any listeners. you should also check this here, otherwise your test is useless

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+
+        // Mock necessary methods
+        Mockito.when(request.method()).thenReturn(HttpMethod.GET);
+        Mockito.when(request.uri()).thenReturn("hookRootURI/registrations/listeners/");
+        Mockito.when(request.getParam("q")).thenReturn(null); // No query param
+        Mockito.when(request.response()).thenReturn(response);
+        Mockito.when(request.headers()).thenReturn(MultiMap.caseInsensitiveMultiMap());
+
+
+        HttpClient selfClient = Mockito.mock(HttpClient.class);
```
mcweba, line 758: Your test should never reach this part of the code (where the HttpClient is used) when you don't provide a search query

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+                            context.async().complete();
+                        })
+                        .onFailure(context::fail);
+            } else {
+                context.fail(ar.cause());
+            }
+        });
+    }
+    @Test
+    public void testSearchListenerNotFound_MultipleListeners_Failure(TestContext context) {
```
mcweba, line 880: Migrate/remove this test to `gateleen-test` module since we don't do integration tests here

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+                            context.async().complete();
+                        })
+                        .onFailure(context::fail);
+            } else {
+                context.fail(ar.cause());
+            }
+        });
+    }
+    @Test
+    public void testSearchListenerNotFound_NoListeners_Failure(TestContext context) {
```
mcweba, line 934: Migrate/remove this test to `gateleen-test` module since we don't do integration tests here

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +942,101 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+    public void testHookHandleSearch_ListenerPathWithValidQueryParam(TestContext context) {
+        Async async = context.async();
+        delete();
+        initRoutingRules();
+
+        String queryParam = "testQuery";
+        String listenerPath = "/_hooks/listeners";
+        String requestUrl = requestUrlBase + listenerPath + "?q=" + queryParam;
+
+        // Register a listener
```
mcweba, line 961: You did not register a listener correctly. Your uri is: 
```
/playground/server/tests/gateleen/monitoredresource/_hooks/listeners
```

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +942,101 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+        TestUtils.registerListener(requestUrlBase + listenerPath, targetUrlBase, new String[]{"GET", "POST"}, null);
+
+        // Send GET request
+        given().queryParam("q", queryParam)
```
mcweba, line 965: Your test creates a GET request to:
```
/playground/server/tests/gateleen/monitoredresource/_hooks/listeners?q=testQuery&q=testQuery
```

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +942,101 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+                .when().get(requestUrl)
+                .then().assertThat().statusCode(200);
+
+        // Validate the response
+        checkGETStatusCodeWithAwait(requestUrl, 200);
```
mcweba, line 970: Your test should verify the content of the response

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +942,101 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+        delete();
+        initRoutingRules();
+
+        String nonMatchingQueryParam = "nonMatchingQuery";
+        String listenerPath = "/_hooks/listeners";
+        String requestUrl = requestUrlBase + listenerPath + "?q=" + nonMatchingQueryParam;
+
+        // Register a listener with a different query param
+        String differentQueryParam = "differentQuery";
+        TestUtils.registerListener(requestUrlBase + listenerPath + "?q=" + differentQueryParam, targetUrlBase, new String[]{"GET", "POST"}, null);
```
mcweba, line 994: You are using query parameter in PUT request for adding listener. This is not correct when register listeners. In this test class there are plenty of examples for registering listeners

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +942,101 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+        initRoutingRules();
+
+        String queryParam = "someQuery";
+        String listenerPath = "/_hooks/listeners";
+        String requestUrl = requestUrlBase + listenerPath + "?q=" + queryParam;
+
+        // No listeners registered
+
+        // Send GET request with a query param
+        given().queryParam("q", queryParam)
```
mcweba, line 1030: Again double query params

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+
+        String queryParam = "testQuery";
+        String routePath = "/routes";
+        String requestUrl = requestUrlBase + routePath + "?q=" + queryParam;
+
+        // Register a route
+        TestUtils.registerRoute(requestUrlBase + routePath, targetUrlBase, new String[]{"GET", "POST"}, null, true, true);
+
+        // Send GET request with a valid query param
+        given().queryParam("q", queryParam)
```
mcweba, line 244: Duplicate query param

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+                .when().get(requestUrl)
+                .then().assertThat().statusCode(200);
+
+        // Validate response
+        checkGETStatusCodeWithAwait(requestUrl, 200);
```
mcweba, line 249: Check for response content

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+        delete();
+        initSettings();
+
+        String nonMatchingQueryParam = "nonMatchingQuery";
+        String routePath = "/routes";
+        String requestUrl = requestUrlBase + routePath + "?q=" + nonMatchingQueryParam;
+
+        // Register a route with a different query param
+        String differentQueryParam = "differentQuery";
+        TestUtils.registerRoute(requestUrlBase + routePath + "?q=" + differentQueryParam, targetUrlBase, new String[]{"GET", "POST"}, null, true, true);
```
mcweba, line 271: Query param in PUT request for route registration. This does not work

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+
+        // Send GET request with non-matching query param
+        given().queryParam("q", nonMatchingQueryParam)
```
mcweba, line 274: Invalid uri

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+                .when().get(requestUrl)
+                .then().assertThat().statusCode(200)
+                .body("routes", Matchers.empty());
+
+        // Validate response
+        checkGETStatusCodeWithAwait(requestUrl, 200);
```
mcweba, line 280: Verify response

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+        initSettings();
+
+        String queryParam = "someQuery";
+        String routePath = "/routes";
+        String requestUrl = requestUrlBase + routePath + "?q=" + queryParam;
+
+        // No routes registered
+
+        // Send GET request with a query param
+        given().queryParam("q", queryParam)
```
mcweba, line 303: invalid uri

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -223,5 +224,91 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+                .when().get(requestUrl)
+                .then().assertThat().statusCode(200)
+                .body("routes", Matchers.empty());
+
+        // Validate response
+        checkGETStatusCodeWithAwait(requestUrl, 200);
```
mcweba, line 309: Verify response

mcweba (changes requested):

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok, I will remove it.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Ok, I will implement this validation

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Ok, I'll use the function already existing in the class

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Added check for response

gateleen-hook/README_hook.md
```diff
@@ -240,10 +240,43 @@ hookHandler.enableResourceLogging(true);
...
 
+## Query-Based Listener and Route Search
 
+Gateleen allows searching for listeners and routes using the query parameter `q`. This simplifies filtering the registered hooks based on query parameters.
 
+### Listener Search with `q`
+Search for listeners based on a query parameter like this:
 
+```
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners/listener/1?q=testQuery
```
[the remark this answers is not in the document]
steniobhz: Fixed

gateleen-hook/README_hook.md
steniobhz: Fixed

gateleen-hook/README_hook.md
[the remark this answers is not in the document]
mcweba: This is not resolved. You did not change the documentation!

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
mcweba: Not resolved, the uri is not fixed!

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -300,15 +306,20 @@ public void testRouteListing_NoRoutesRegistered(TestContext context) {
...
-                .body("routes", Matchers.empty());
+                .extract().response();
+
+        // Print the body of the response for debugging
+        System.out.println("Response body: " + response.getBody().asString());
 
-        // Validate response
-        checkGETStatusCodeWithAwait(requestUrl, 200);
+        // Assert that the response body is empty or does not contain routes
+        Assert.assertTrue("No routes should be registered",
```
mcweba, line 318: What do you assert here? Your test will be green for outputs like

```json
{"routes":[]}
```
but also for
```json
{"routes":["foo","bar"]}
```
and that is probably not what you want to assert here

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Fixed

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -300,15 +306,20 @@ public void testRouteListing_NoRoutesRegistered(TestContext context) {
...
         // Send GET request with a query param
-        given().queryParam("q", queryParam)
+        Response response = given().queryParam("q", queryParam)
                 .when().get(requestUrl)
                 .then().assertThat().statusCode(200)
-                .body("routes", Matchers.empty());
+                .extract().response();
+
+        // Print the body of the response for debugging
+        System.out.println("Response body: " + response.getBody().asString());
```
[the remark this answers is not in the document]
steniobhz: Removed!

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Created a new validate to check it

gateleen-hook/README_hook.md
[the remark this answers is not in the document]
steniobhz: Fixed

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Fixed, Implemented using existing examples in other tests

gateleen-hook/README_hook.md
```diff
@@ -248,7 +248,7 @@ Gateleen allows searching for listeners and routes using the query parameter `q`
 Search for listeners based on a query parameter like this:
 
 ```
-GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners/listener/1?q=testQuery
+GET http://myserver:7012/gateleen/server/listenertest/_hooks/listeners?q=testQuery
```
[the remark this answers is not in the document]
steniobhz: Running locally it was possible to list using
http://localhost:7012/listeners?q=test
or in the case of routes
http://localhost:7012/routes?q=test
This is because the request arrives at handle, the currently validation checks only if contains the LISTENERS_KEY or the ROUTES_KEY (is that correct?)

Anyway, I updated the suggested url.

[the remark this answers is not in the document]
mcweba: > Running locally it was possible to list using http://localhost:7012/listeners?q=test or in the case of routes http://localhost:7012/routes?q=test This is because the request arrives at handle, the currently validation checks only if contains the LISTENERS_KEY or the ROUTES_KEY (is that correct?)
> 
> Anyway, I updated the suggested url.

No this is not correct. Please read my remarks. I already commented that it's wrong to rely on "contains 'listeners'" only.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Validation has been implemented as suggested.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: I changed it to status code instead of putting 400 directly

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: I agree, implement the solution as instructed

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+        boolean handled = hookHandler.handle(routingContext);
+
+        // Assert
+        context.assertFalse(handled);
+        Mockito.verify(response, Mockito.never()).end();
+    }
+
+
+    @Test
+    public void testSearchMultipleListeners_Success(TestContext context) {
```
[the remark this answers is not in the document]
steniobhz: I moved the test to the tests module

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -641,6 +643,333 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+                            context.async().complete();
+                        })
+                        .onFailure(context::fail);
+            } else {
+                context.fail(ar.cause());
+            }
+        });
+    }
+    @Test
+    public void testSearchSingleListener_Success(TestContext context) {
```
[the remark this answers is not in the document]
steniobhz: I moved the test to the tests module

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: I moved the test to the tests module

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: I moved the test to the tests module

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: I added an extra check for the response.

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: fixed

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
steniobhz: Since it was an integration test, it was moved to the integration testing module.

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
```diff
@@ -224,4 +226,112 @@ private void assertResponse(final Response response, final String[] expectedArra
...
+     * Test for route listing with a valid query parameter.
+     */
+    @Test
+    public void testRouteListing_ValidQueryParam(TestContext context) {
+        Async async = context.async();
+        delete(); // Remove any pre-existing data
+        initSettings(); // Initialize routing rules
+
+        String queryParam = "routeTests";
+        String routePath = "/routes";
```
mcweba, line 239: Variable `routePath` is not used

mcweba (changes requested):
There seems to be an issue in the implementation of route searching. When listing the routes with
```

GET /playground/server/hooks/v1/registrations/routes
```
you get the following response
```json

{
    "routes": [
        "+playground+server+destination1+_hooks+routes+http+myhook54"
    ]
}
```
However when you add the search query parameter like this
```
GET /playground/server/hooks/v1/registrations/routes?q=r1
```
you get the following response
```json
{
    "routes": [
        "/playground/server/destination1"
    ]
}
```
The response should still be the same. This is the case for listeners hooks but not for route hooks.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
mcweba: Added another remark about this, since it is not yet fixed.

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
mcweba: Where is the new validation to check the repsonse? This is still not fixed since my statements about the possible responses that satisfy your validation is still true.

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: removed

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+
+        // Verify that the listener was correctly registered
+        Response response = given()
+                .queryParam("q", listenerName)
+                .when().get(searchUrlBase)
+                .then().assertThat().statusCode(200)
+                .extract().response();
+
+        // Assert that the response contains the expected query param
+        System.out.println(response.getBody().asString());
```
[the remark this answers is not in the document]
steniobhz: Removed

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+
+        // Verify that the listener search with a no registered listener
+        Response response = given()
+                .queryParam("q", queryParam)
+                .when().get(searchUrlBase)
+                .then().assertThat().statusCode(200)
+                .extract().response();
+
+        // Assert that the response contains the expected query param
+        Assert.assertFalse(response.getBody().asString().contains(queryParam)); // Fails if not found
```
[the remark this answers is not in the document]
steniobhz: Added

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+
+
+    @Test
+    public void testHookHandleSearch_ListenerPathInvalidParam(TestContext context) {
+        Async async = context.async();
+        delete();
+        initRoutingRules();
+
+        String queryParam = "testQuery";
+        String listenerPath = "/_hooks/listeners";
```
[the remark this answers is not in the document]
steniobhz: removed

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+        String requestUrl = searchUrlBase+ "?www=" + queryParam;
```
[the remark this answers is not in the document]
steniobhz: Created

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+        checkGETStatusCodeWithAwait(requestUrl, 400);
+        async.complete();
+    }
+
+    /**
+     * Test for hookHandleSearch with listener storage path and no query parameter. <br />
+     * requestUrl: http://localhost:7012/playground/server/hooks/v1/registrations/listeners/?q=
+     */
+    @Test
+    public void testHookHandleSearch_NoQueryParameter(TestContext context) {
```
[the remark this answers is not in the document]
steniobhz: renamed

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
steniobhz: Implemented other validations to cover all situations.

gateleen-test/src/test/java/org/swisspush/gateleen/hook/RouteListingTest.java
steniobhz: Add validation to check if contains the value and also check if the result is empty

gateleen-hook/README_hook.md
```diff
@@ -240,10 +240,40 @@ hookHandler.enableResourceLogging(true);
 ```
 
 
+## Query-Based Listener and Route Search
 
+Gateleen allows searching for listeners and routes using the query parameter `q`. This simplifies filtering the registered hooks based on query parameters.
```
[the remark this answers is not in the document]
steniobhz: Add the information that "The search will be based on the value registred of the destination property"

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -592,6 +605,63 @@ public boolean handle(final RoutingContext ctx) {
...
+     * @param repository The items to search .
+     * @param getDestination Function to extract destinations.
+     * @param queryParam The query string to match.
+     * @param resultKey The key for the result in the response.
+     * @param response The HTTP response to return the results. Must not be null.
+     */
+    private <T> void handleSearch(Map<String, T> repository, Function<T, String> getDestination, String queryParam, String resultKey, HttpServerResponse response) {
+
+        if (queryParam == null || queryParam.isEmpty()) {
+            response.setStatusCode(StatusCode.BAD_REQUEST.getStatusCode()).end("Bad Request: One or more required parameters are missing or null");
```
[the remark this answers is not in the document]
steniobhz: Added the validation to fix it

gateleen-hook/src/test/java/org/swisspush/gateleen/hook/HookHandlerTest.java
```diff
@@ -642,6 +661,188 @@ public void hookRegistration_RouteWithRouteMultiplier(TestContext testContext) t
...
+        // Validate JSON response content for matching listener
+        String jsonResponse = responseCaptor.getValue();
+        assertNotNull(jsonResponse);
+        assertTrue(jsonResponse.contains(singleListener));
+    }
+
+    @Test
+    public void testHandleGETRequestWithListenersSearchMultipleResults() throws InterruptedException {
+        // Define the URI and set up the GET request with a broader search parameter for multiple listeners
+        String uri = "/hookRootURI/registrations/listeners";
```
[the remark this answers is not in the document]
steniobhz: HOOK_LISTENER_URI created to use for all this cases

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -942,4 +947,233 @@ private void checkGETStatusCodeWithAwait(final String request, final Integer sta
...
+     * requestUrl: http://localhost:7012/playground/server/hooks/v1/registrations/listeners/?q=someQuery
+     */
+    @Test
+    public void testHookHandleSearch_NoListenersRegistered(TestContext context) {
+        Async async = context.async();
+        delete();
+        initRoutingRules();
+
+        String queryParam = "someQuery";
+        String listenerPath = "/hooks/listeners";
```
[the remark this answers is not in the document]
steniobhz: Removed!

gateleen-test/src/test/java/org/swisspush/gateleen/hook/ListenerTest.java
```diff
@@ -54,6 +57,8 @@ public void initRestAssured() {
 
         requestUrlBase = "/tests/gateleen/monitoredresource";
         targetUrlBase = "http://localhost:" + MAIN_PORT + SERVER_ROOT + "/tests/gateleen/targetresource";
+        searchUrlBase = "http://localhost:" + MAIN_PORT + SERVER_ROOT + "/hooks/v1/registrations/listeners";
```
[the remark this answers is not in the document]
steniobhz: testHookHandleSearch_UseSlashInSearchUrlBase created to cover this.

mcweba (approved):

PR #620, opened 2024-12-18
Bug 618 response handler in forwarder called twice per request

hoetdr-post:
On line 316, just after creating it, the cResHandler is added as a response handler to the cReq. No need to add it again when sending the request.
At first at removed it on line 316, since we add in while sending, but that doesn't work, because in some cases the pump starts writing the request before the send method is called and if the server already answers before the send is called, we would have no response handler installed.
This happens when forwarding to /server/return-with-status-code/503


hiddenalpha (approved):
After TelCo I think this change is very valid.

PR #627, opened 2025-01-08
#654 collecting kafka metrics

mcweba:


gateleen-kafka/src/main/java/org/swisspush/gateleen/kafka/KafkaMessageSender.java
```diff
@@ -44,7 +63,45 @@ private Future<Void> sendMessage(KafkaProducer<String, String> kafkaProducer, Ka
...
+
+        if(meterRegistry != null) {
+            Counter newCounter = Counter.builder(SUCCESS_SEND_MESSAGES_METRIC)
+                    .description(SUCCESS_SEND_MESSAGES_METRIC_DESCRIPTION)
+                    .tag(TOPIC, topic)
+                    .register(meterRegistry);
+            newCounter.increment();
+            successSendCounterMap.put(topic, newCounter);
+        }
+    }
```
hiddenalpha, line 89: `incrementSuccessCount` and `incrementFailCount1` are NOT thread-save. Is it ensured only one thread is at work here (eg in case of multiple verticle instances)?

hiddenalpha (approved):

Issue #653, opened 2025-05-15
Store delta values to the correct storage


mcweba:
With the current implementation of the [gateleen-delta](https://github.com/swisspost/gateleen/tree/develop/gateleen-delta) module, the delta values are always stored in the main storage, even if the resource is for example stored in _add_1_ storage.

This can cause problems because when a resource is deleted, also the delta values are deleted. In cases where a resource from an additional storage is deleted, the delta values are not deleted (because they reside in another storage and cannot be found in the lua script).

Also when the delta values don't expire (when no x-expire-after request header was present on writing), they will stay in the storage forever and cannot be removed easily.

mcweba:
Released in [v2.1.21](https://github.com/swisspost/gateleen/releases/tag/v2.1.21)

PR #654, opened 2025-05-20
#653 extend RedisProvider to define storage name

mcweba:


gateleen-core/src/main/java/org/swisspush/gateleen/core/redis/RedisProvider.java
```diff
@@ -11,4 +13,18 @@
...
+     * This default implementation ignores the provided storage name
+     * and simply returns the default RedisAPI instance.
+     * </p>
+     *
+     * @param storageName the name of the storage (can be null)
+     * @return a Future containing the RedisAPI instance
+     */
+    default Future<RedisAPI> redis(@Nullable String storageName) {
+        return redis();
+    }
```
hiddenalpha, line 29: (**WARN: opinion ahead**)

```java
    Future<RedisAPI> redis(@Nullable String storageName);
```
This design decision looks really awkward to me.

IMHO a `RedisProvider` should be responsible for exactly one instance (*not a set of named instances*). This is the use-case 99% of all usage code needs.

So whenever any code needs several redis, it could express this for example as something like:
```java
    IamAClassThatNeedsMultipleRedis(Function<String, Future<RedisAPI> getRedisByName);
```
Or if it really needs the provider and not the redis itself:
```java
    IamAClassThatNeedsMultipleRedis(Function<String, RedisProvider> getRedisByName);
```
Or maybe we even introduce another (**completely independent, eg NO inheritance relation to the existing RedisProvider**!):
```java
    public interface RedisByNameProvider {
        Future<RedisAPI> redisByName(String name);
    }
    ...
    IamAClassThatNeedsMultipleRedis(RedisByNameProvider redisProvider);
```

References:

For those who prefer more academic reads, consider research about `InterfaceSegregationPrinciple` from SOLID.

Issue #663, opened 2025-06-18
Cleanup unwanted properties in user profiles


mcweba:
The class handling user profiles is called `UserProfileHandler`. This class accepts a `UserProfileConfiguration` holding a list of `allowedProfileProperties`.

Whenever a user profile is updated, only the properties defined in the `allowedProfileProperties` list are stored in the storage. However, when the property is already in the storage, it can never be removed because the PUT request to the storage is called with a **?merge=true** parameter.

Because of this parameter, it is not possible to clean the user profile from properties which are not wanted anymore.

PR #665, opened 2025-06-24
#663 Add option removeNotAllowedProfileProperties to clean unwanted p…

mcweba:
…roperties from user profiles

### DIFF
--- a/gateleen-user/src/main/java/org/swisspush/gateleen/user/UserProfileHandler.java
+++ b/gateleen-user/src/main/java/org/swisspush/gateleen/user/UserProfileHandler.java
@@ -171,10 +171,10 @@ public class UserProfileHandler implements LoggableResource {
         Set<String> profileFieldNames = profileCopy.fieldNames();
         profileFieldNames.stream().filter(fieldName -> !userProfileConfiguration.isAllowedProfileProperty(fieldName)).forEach(fieldName -> {
             if(userProfileConfiguration.isRemoveNotAllowedProfileProperties()) {
-                log.debug("Removing property '{}' from user profile", fieldName);
+                log.debug("Property '{}' is not allowed, removing it from user profile", fieldName);
                 profile.put(fieldName, null);
             } else {
-                log.debug("Property '{}' is not allowed, but because of the `merge` option not removed from user profile", fieldName);
+                log.debug("Property '{}' is not allowed, removing it from user profile (only when not already in storage)", fieldName);
                 profile.remove(fieldName);
             }
         });


gateleen-user/src/main/java/org/swisspush/gateleen/user/UserProfileHandler.java
```diff
@@ -186,8 +170,13 @@ protected void cleanupUserProfile(final JsonObject profile, final Handler<JsonOb
...
         Set<String> profileFieldNames = profileCopy.fieldNames();
         profileFieldNames.stream().filter(fieldName -> !userProfileConfiguration.isAllowedProfileProperty(fieldName)).forEach(fieldName -> {
-            log.debug("Removing property '{}' from user profile", fieldName);
-            profile.remove(fieldName);
+            if(userProfileConfiguration.isRemoveNotAllowedProfileProperties()) {
+                log.debug("Removing property '{}' from user profile", fieldName);
+                profile.put(fieldName, null);
+            } else {
+                log.debug("Property '{}' is not allowed, but because of the `merge` option not removed from user profile", fieldName);
+                profile.remove(fieldName);
```
hiddenalpha, line 178: Aren't those two statements contradicting? Log says we do NOT remove, but then code literally calls `remove`?
```java
    log ... not removed ... ;
    profile.remove(fieldName);
```

gateleen-user/src/main/java/org/swisspush/gateleen/user/UserProfileHandler.java
```diff
@@ -186,8 +170,13 @@ protected void cleanupUserProfile(final JsonObject profile, final Handler<JsonOb
         JsonObject profileCopy = profile.copy();
         Set<String> profileFieldNames = profileCopy.fieldNames();
         profileFieldNames.stream().filter(fieldName -> !userProfileConfiguration.isAllowedProfileProperty(fieldName)).forEach(fieldName -> {
-            log.debug("Removing property '{}' from user profile", fieldName);
-            profile.remove(fieldName);
+            if(userProfileConfiguration.isRemoveNotAllowedProfileProperties()) {
+                log.debug("Removing property '{}' from user profile", fieldName);
+                profile.put(fieldName, null);
```
hiddenalpha, line 175: I assume there's a reason we do `put(fieldName, null)` in place of `remove(fieldName)`?

mcweba: Yes the reason is, that `remove(fieldName)` does not actually remove a field from storage when it's already there. This is because of the `merge=true` param that is used when the storage is called.

PR #668, opened 2025-07-01
fix unhandled exception

ZhengXinCN:
io.vertx.core.json.DecodeException: Failed to decode:String value length (20054016) exceeds the maximum allowed (20000000, from StreamReadConstraints.getMaxStringLength())
	at io.vertx.core.json.jackson.DatabindCodec.fromParser(DatabindCodec.java:129) ~[vertx-core-4.5.1.jar:4.5.1]
	at io.vertx.core.json.jackson.DatabindCodec.fromString(DatabindCodec.java:90) ~[vertx-core-4.5.1.jar:4.5.1]
	at io.vertx.core.json.JsonObject.fromJson(JsonObject.java:1250) ~[vertx-core-4.5.1.jar:4.5.1]
	at io.vertx.core.json.JsonObject.<init>(JsonObject.java:51) ~[vertx-core-4.5.1.jar:4.5.1]


mcweba (changes requested):
Could you add a unit test showing what the actual problem is? Do you get the DecodeException when working with too large json resources?

ZhengXinCN:
@mcweba I see many such exception in the Internal project from logs. a test also added, but there is no way to easy test this, so I extracted the key part of code into a function.

mcweba (approved):

PR #708, opened 2025-11-10
Trackable broadcast

ZhengXinCN:
log gateleen key event consumer counter

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/event/TrackableEventPublish.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/event/TrackableEventPublish.java
@@ -112,11 +112,13 @@ public class TrackableEventPublish {
             if (trackerEnabled) {
                 if (msg.body() instanceof JsonObject) {
                     JsonObject body = (JsonObject) msg.body();
+                    // we may don't have replay address if the function just switch ON.
+                    // the event before the switch on may just arrive
                     if (body.containsKey(KEY_REPLY_ADDRESS)) {
                         String replyAddr = body.getString(KEY_REPLY_ADDRESS);
                         vertx.eventBus().send(replyAddr, null);
-                        handler.handle(body.getString(KEY_DATA));
                     }
+                    handler.handle(body.getString(KEY_DATA));
                 }
             } else {
                 if (msg.body() instanceof JsonObject) {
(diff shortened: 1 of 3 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1238,6 +1239,7 @@ private void handleListenerRegistration(final HttpServerRequest request) {
                         RequestLogger.logRequest(vertx.eventBus(), request, status, buffer);
                     }
                     vertx.eventBus().publish(SAVE_LISTENER_ADDRESS, listenerStorageUri);
+                    trackableEventPublish.publish(SAVE_LISTENER_ADDRESS, listenerStorageUri);
```
hiddenalpha: I assume it is by intent to DUPLICATE that publication here for whatever reason? All other places do a REPLACE of the call. Except this one, which now does TWO calls.
```patch
     vertx.eventBus().publish(SAVE_LISTENER_ADDRESS, listenerStorageUri);
+    trackableEventPublish.publish(SAVE_LISTENER_ADDRESS, listenerStorageUri);
```

ZhengXinCN: Thanks, I removed it already, but don't know it doesn't marked it as outdated

gateleen-core/src/main/java/org/swisspush/gateleen/core/event/TrackableEventPublish.java
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public void consumer(String address, Handler<String> handler) {
+        vertx.eventBus().consumer(address, msg -> {
+            if (trackerEnabled) {
+                if (msg.body() instanceof JsonObject) {
+                    JsonObject body = (JsonObject) msg.body();
+                    if (body.containsKey(KEY_REPLY_ADDRESS)) {
+                        String replyAddr = body.getString(KEY_REPLY_ADDRESS);
+                        vertx.eventBus().send(replyAddr, null);
+                        handler.handle(body.getString(KEY_DATA));
```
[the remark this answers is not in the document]
ZhengXinCN: Yes, there 2 things need to check to use reply:
1. function enabled
2. there is a reply address attached.

because when  trackerEnabled == true, may still received some events which published by other when tracker is not enabled yet

PR #729, opened 2026-02-10
try to improve memory leak in route

ZhengXinCN:
- changed some map to use thread safe one
- remove the meter from meterRegistry of Forwarder, to prevents the object retained
- disable per rule base http client's meter, to prevents the HttpClient retained in memory.
- can not prove how much it will help

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/LocalListenerRepository.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/LocalListenerRepository.java
@@ -11,10 +11,10 @@ import java.util.regex.Pattern;
 
 /**
  * Local in-memory implementation of a LocalListenerRepository.
- * 
+ *
  * @author https://github.com/ljucam [Mario Ljuca]
  */
-public class LocalListenerRepository extends ListenerRepositoryBase<Map<String, Set<Listener>>>implements ListenerRepository {
+public class LocalListenerRepository extends ListenerRepositoryBase<Map<String, Set<Listener>>> implements ListenerRepository {
     private Logger log = LoggerFactory.getLogger(LocalListenerRepository.class);
 
     /*
@@ -58,7 +58,7 @@ public class LocalListenerRepository extends ListenerRepositoryBase<Map<String, 
             return listeners;
         });
 
-        listenerToUrlMap.put(listener.getListenerId(), listener);
+        listenerToUrlMap.compute(listener.getListenerId(), (s, existListener) -> listener);
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/LocalListenerRepository.java
```diff
@@ -30,8 +31,8 @@ public class LocalListenerRepository extends ListenerRepositoryBase<Map<String,
      * Creates a new instance of the local in-memory LocalHookListenerRepository.
      */
     public LocalListenerRepository() {
-        urlToListenersMap = new HashMap<>();
-        listenerToUrlMap = new HashMap<>();
+        urlToListenersMap = new ConcurrentHashMap<>();
+        listenerToUrlMap = new ConcurrentHashMap<>();
```
hiddenalpha: **WARN**: Method `removeListener()` performs **non-atomic updates** on `listenerToUrlMap`. Just using a thread-safe map does not solve this concurrency issue there.

hiddenalpha: **WARN**: Method `addListener()` updates both of `urlToListenersMap` and `listenerToUrlMap` with the possibility of other threads modifying one of the maps before the 2nd update can be done.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -158,6 +158,12 @@ public void setMeterRegistry(MeterRegistry meterRegistry) {
         }
     }
 
+    public void shutdown() {
+        if (this.meterRegistry != null) {
+            meterRegistry.remove(forwardTimer);
```
mcweba: Do not remove the timer from the registry as this can cause gaps in the graphs, alerts firing or resetting and confusing "where did me metric go?" moments. Since timers are quite static in the sense that only on routing rule change new instances are created, there is no serious risk for memory leaks.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
```diff
@@ -288,7 +288,8 @@ public HttpClientOptions buildHttpClientOptions() {
                 .setKeepAlive(isKeepAlive())
                 .setKeepAliveTimeout(getKeepAliveTimeout())
                 .setPipelining(false)
-                .setMaxWaitQueueSize(getMaxWaitQueueSize());
+                .setMaxWaitQueueSize(getMaxWaitQueueSize())
+                .setMetricsName(null);
```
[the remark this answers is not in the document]
ZhengXinCN: I still not sure about this yet, did see any related doc, but google told me like that
image

ZhengXinCN:
> Where did you find memory leaks in routes? Can you provoke them?

we have 157x route, but you see how many instance here

image

mcweba:
> > Where did you find memory leaks in routes? Can you provoke them?
> 
> we have 157x route, but you see how many instance here
> 
> image

In fact that's interesting. Why are there so many instances of Hooks and Rules, etc.

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
ZhengXinCN: I updated code to remove reference from forwarder, not remove from registry.
from my local test case, Forwarder can be removed heap, no leak.

ZhengXinCN:
@hiddenalpha , where I can see those warnings, I try to fix them, but can't find where I can see results. Thanks

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Rule.java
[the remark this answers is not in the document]
ZhengXinCN: Tested, not work. reverted

PR #742, opened 2026-03-30
allow to decode base64 legacy format

ZhengXinCN:
because we still have old data (created before legacy mode off) in the database, so need a function can handle the legacy mode base64 format

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/event/EventBusHandler.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/event/EventBusHandler.java
@@ -24,6 +24,7 @@ import org.swisspush.gateleen.core.configuration.ConfigurationResourceConsumer;
 import org.swisspush.gateleen.core.configuration.ConfigurationResourceManager;
 import org.swisspush.gateleen.core.http.RequestLoggerFactory;
 import org.swisspush.gateleen.core.json.JsonMultiMap;
+import org.swisspush.gateleen.core.util.Base64Unit;
 
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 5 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
```diff
@@ -82,8 +83,9 @@ public HttpRequest(JsonObject object) {
         JsonArray headersArray = object.getJsonArray("headers");
         if (headersArray != null) {
             this.headers = JsonMultiMap.fromJson(headersArray);
+            this.payload = object.getBinary("payload");
         }
-        this.payload = object.getBinary("payload");
+        this.payload = Base64Unit.decodeBase64Safe(object.getString("payload"));
```
mcweba, line 87: Why is the payload decoded twice?

gateleen-core/src/main/java/org/swisspush/gateleen/core/http/HttpRequest.java
```diff
@@ -82,8 +83,8 @@ public HttpRequest(JsonObject object) {
         JsonArray headersArray = object.getJsonArray("headers");
         if (headersArray != null) {
             this.headers = JsonMultiMap.fromJson(headersArray);
+            this.payload = Base64Unit.decodeBase64Safe(object.getString("payload"));
```
mcweba: Why should the payload only be decoded when the `headersArray` is not null? Before this call was outside of this if-statement!

gateleen-core/src/main/java/org/swisspush/gateleen/core/event/EventBusHandler.java
```diff
@@ -183,6 +183,8 @@ public boolean handle(final HttpServerRequest request) {
                                         } else if (responseContentType != null && responseContentType.contains(TEXT)) {
                                             rsp.end(response.getString(PAYLOAD));
                                         } else {
+                                            //TODO: remove decodeBase64Safe back to getBinary once we have no more old data in storage
+                                            //rsp.end(Buffer.buffer(response.getBinary(PAYLOAD)));
```
Kusig: Just a question: Wouldn't it be possible to do this somehow backward compatible, meaning trying the new one and if it fails, using the old one as fallback?  
The reason for the question is due to the comment about "no more old data" because from where and when do we   know when we don't have anymore "old data".

ZhengXinCN: yes, it is possible, I will add it

Kusig (approved):

PR #745, opened 2026-04-08
Feature/issue743 mget cluster fix

mcweba:
Implemented a ClusterSafeMget utility (similar to vertx-redisques)

### DIFF
--- a/gateleen-core/src/main/java/org/swisspush/gateleen/core/redis/ClusterSafeMget.java
+++ b/gateleen-core/src/main/java/org/swisspush/gateleen/core/redis/ClusterSafeMget.java
@@ -1,0 +1,259 @@
+package org.swisspush.gateleen.core.redis;
+
+import io.vertx.core.Future;
+import io.vertx.redis.client.RedisAPI;
+import io.vertx.redis.client.Response;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.nio.charset.StandardCharsets;
+import java.util.*;
+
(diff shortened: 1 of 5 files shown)


gateleen-core/src/main/java/org/swisspush/gateleen/core/redis/ClusterSafeMget.java
```diff
@@ -0,0 +1,259 @@
...
+     *       and XOR with the CCITT polynomial {@code 0x1021}; otherwise just shift left.</li>
+     *   <li>Mask to 16 bits with {@code & 0xFFFF} to discard carry bits produced by Java's
+     *       32-bit {@code int} arithmetic.</li>
+     * </ol>
+     * </p>
+     *
+     * @return a 256-element array where {@code table[i]} is the CRC-16/CCITT-FALSE remainder
+     *         for the single-byte input value {@code i}
+     */
+    private static int[] buildCrc16Table() {
```
Kusig: Is there a reason why you don't use the Zmodem method as it was done in Redisques to calculate the Slot?

mcweba: I implemented this before I heard from the zmodem method. I will have a look

mcweba: Changed the implementation to also use ZModem. Thanks for the hint.

Kusig (approved):

ZhengXinCN (approved):

PR #762, opened 2026-06-01
MSOP-7527 fixed the applyQueueConfig may fail if we have duplicated header

ZhengXinCN:
- fixed applyQueueConfig function that failed if we have duplicated header
- move function comment into interface
- added more tests to convers the almost use case

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -1747,7 +1747,7 @@ public class HookHandler implements LoggableResource {
         // try to find X_QUEUE_CONFIG_NAME_PATTERN and X_QUEUE_CONFIG_MAX_LIMIT headers
         for (Object obj : headers) {
             if (!(obj instanceof JsonObject)) {
-                log.info("Ignoring invalid header config entry: {}", obj);
+                log.debug("Ignoring invalid header for per queue config entry: {}", obj);
                 continue;
             }
             JsonObject headerPair = (JsonObject) obj;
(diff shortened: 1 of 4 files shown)


gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -1732,28 +1732,63 @@ private void registerRoute(Buffer buffer) {
...
+            return;
+        }
+
+        String queueNamePattern = null;
+        String maxLimitRaw = null;
+
+        // try to find X_QUEUE_CONFIG_NAME_PATTERN and X_QUEUE_CONFIG_MAX_LIMIT headers
+        for (Object obj : headers) {
+            if (!(obj instanceof JsonObject)) {
+                log.warn("Ignoring invalid header config entry: {}", obj);
```
srudin: I don't know how the logging in Gateleen is generally handled. IMO this is not a warning - a warning is a situation that requires somebody to check something. I don't think that's the case here so I have the feeling that this and the other log messages in this PR will spam the log with no real benefit, instead it makes it harder to find warnings that really require attention.
But like I said, I don't know the logging philosophy in Gateleen so if this is the way it's done here, fine for me.

srudin (approved):

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
ZhengXinCN: I changed to DEBUG level, because  those invalid are all normal, just not a queue limit config.

Kusig (approved):

PR #768, opened 2026-06-05
Sdcisa 15631 fixed possible memory leak

ZhengXinCN:
fixed possible memory leak:
- use cache to expires a QueueSplitExecutor in case long time not in use
- removed all Dynamic QueueSplitExecutors when config resource removed
- fixed the wrong header will skip the QueueSplitExecutor exist check
- more tests....

### DIFF
--- a/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
+++ b/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
@@ -1,6 +1,8 @@
 package org.swisspush.gateleen.queue.queuing.splitter;
 
 import com.google.common.base.Strings;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
 import io.vertx.core.buffer.Buffer;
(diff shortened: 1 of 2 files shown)


gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
```diff
@@ -82,38 +89,42 @@ private void initializeQueueSplitterConfiguration(Buffer configuration) {
...
+        final QueueSplitExecutorFromStaticList existDynamicQueueSplitExecutor = dynamicQueueSplitExecutors.getIfPresent(queueName);
+
+        if (Strings.isNullOrEmpty(numberOfQueueString) && existDynamicQueueSplitExecutor == null) {
             // do nothing
             return queueName;
         }
 
-        int numberOfQueue;
+        // take the numberOfQueue from exist executor, if there is one.
+        int numberOfQueue = existDynamicQueueSplitExecutor == null ? 0 : existDynamicQueueSplitExecutor.getConfiguration().getPostfixFromStatic().size();
```
mcweba: Produces _NullPointerException_ when `getPostfixFromStatic()` returns `null`

mcweba (changes requested):

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/splitter/QueueSplitterImpl.java
ZhengXinCN: I will added the check, for prevent some changes in future but didn't take care about this

mcweba:
There was no commit after the last review. What should I review now?

ZhengXinCN:
> There was no commit after the last review. What should I review now?
sorry forget check does push done, happened many times
image

mcweba (approved):

Kusig (approved):

PR #769, opened 2026-06-17
SDCISA-22735 the enqueue error from redisques was ignored

ZhengXinCN:
surprised that the enqueue error from redisques side was dropped, and return a OK (200), this should not happens, because we lost data
The new solution is pass a flag into Before Handler, and reject server request there

### DIFF
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -945,7 +945,7 @@ public class HookHandler implements LoggableResource {
         // is executed. This way the after handler will
         // also be called properly.
         if (filteredListeners.isEmpty() && handler != null) {
-            handler.handle(null);
+            handler.handle(true);
         }
     }
 
--- a/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
+++ b/gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
@@ -846,7 +846,7 @@ public class HookHandler implements LoggableResource {
 
             // Create handlers for before/after - cases
             Handler<Void> afterHandler = installAfterHandler(ctx, buffer, afterListener);
-            Handler<Void> beforeHandler = installBeforeHandler(ctx, buffer, beforeListener, afterHandler);
+            Handler<Boolean> beforeHandler = installBeforeHandler(ctx, buffer, beforeListener, afterHandler);
 
             // call the listeners (before)
             callListener(ctx, buffer, beforeListener, beforeHandler);
@@ -861,7 +861,7 @@ public class HookHandler implements LoggableResource {
      * @param filteredListeners all listeners which should be called
      * @param handler           the handler, which should handle the requests
      */
-    private void callListener(RoutingContext ctx, final Buffer buffer, final List<Listener> filteredListeners, final Handler<Void> handler) {
+    private void callListener(RoutingContext ctx, final Buffer buffer, final List<Listener> filteredListeners, final Handler<Boolean> handler) {
         HttpServerRequest request = ctx.request();
         for (Listener listener : filteredListeners) {
             log.debug("Enqueue request matching {} {} with listener {}", request.method(), listener.getMonitoredUrl(), listener.getListener());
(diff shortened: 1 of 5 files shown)


gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
```diff
@@ -242,24 +242,29 @@ void enqueue(final HttpServerRequest request, HttpRequest queuedRequest, final S
...
                                 monitoringHandler.updateLastUsedQueueSizeInformation(queue);
                                 monitoringHandler.updateEnqueue();
                             }
 
                             if (request != null) {
-                                ResponseStatusCodeLogUtil.info(request, StatusCode.ACCEPTED, QueueClient.class);
-                                request.response().setStatusCode(StatusCode.ACCEPTED.getStatusCode());
-                                request.response().setStatusMessage(StatusCode.ACCEPTED.getStatusMessage());
+                                ResponseStatusCodeLogUtil.info(request, StatusCode.OK, QueueClient.class);
+                                request.response().setStatusCode(StatusCode.OK.getStatusCode());
```
ZhengXinCN: All reviewers: I am not sure about this, but we can see in old case, there no ServerRequest pass in, will response a 200, and if there is a ServerRequest passed in will response a 202. and problem is here in case no ServerRequest and the enqueue failed, it returns a 200 as default

Kusig: What do you mean by ServerRequest, there are "ClientReqeusts" and "SelfRequests" according to the doc.?

We do change here the only the StatusCode and therefore we break potential clients or logic based on ACCEPTED.    So there is care needed here, isn't it?

I still don't get why the former ACCEPTED is a problem here at all.  A client should deal with 200 or 202 the same way?

mcweba: You cannot just replace the `202 Accepted` response with '200 OK'. Successful enqueues must always return a `202 Accepted`. A failing enqueue without ServerRequest should not return '200 OK', this would actually be a bug

mcweba (changes requested):

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
ZhengXinCN: Yes, I found many thing failed after this change. I already find a new way.

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
```diff
@@ -108,18 +108,20 @@ public void enqueue(HttpRequest request, final String queue, final Handler<Void>
...
                     if (OK.equals(event.result().body().getString(STATUS)) && monitoringHandler != null) {
+                        succeed = true;
                         monitoringHandler.updateLastUsedQueueSizeInformation(queue);
                         monitoringHandler.updateEnqueue();
                     }
                     // call the done handler to tell,
                     // that the request was written
                     if (doneHandler != null) {
-                        doneHandler.handle(null);
+                        doneHandler.handle(succeed);
```
ZhengXinCN, line 124: tell before handler what is enqueue result

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
```diff
@@ -255,18 +258,21 @@ void enqueue(final HttpServerRequest request, HttpRequest queuedRequest, final S
...
+                                request.response().end(event.result().body().getString(MESSAGE));
+                            }
                         }
                     }
 
                     // call the done handler to tell,
                     // that the request was written
                     if (doneHandler != null) {
-                        doneHandler.handle(null);
+                        doneHandler.handle(succeed);
```
ZhengXinCN, line 275: tell before handler what is enqueue result

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -976,46 +976,53 @@ private Handler<Void> installAfterHandler(final RoutingContext ctx, final Buffer
...
+                             * enqueueing process, therefore it is
+                             * imperative to use isRequestAlreadyHooked(HttpServerRequest request)
+                             * before calling the handle method of
+                             * this class!
+                             */
+                            createSelfRequest(ctx.request(), buffer, afterHandler);
+                        }
                     }
+                }else {
+                    log.warn("Request failed: {}", ctx.request().uri());
```
ZhengXinCN, line 1022: enqueue failed, just reject the request

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
```diff
@@ -226,14 +228,15 @@ private void enqueue(final HttpServerRequest request, HttpRequest queuedRequest,
      * @param queue         queue
      * @param doneHandler   a handler which is called as soon as the request is written into the queue.
      */
-    void enqueue(final HttpServerRequest request, HttpRequest queuedRequest, final String queue, final Handler<Void> doneHandler) {
+    void enqueue(final HttpServerRequest request, HttpRequest queuedRequest, final String queue, final Handler<Boolean> doneHandler) {
         if (!QueueProcessor.httpMethodIsQueueable(queuedRequest.getMethod())) {
             log.warn("Ignore enqueue of unsupported HTTP method in '{} {}'.", queuedRequest.getMethod(), queuedRequest.getUri());
             if (doneHandler != null) doneHandler.handle(null);
```
mcweba: NullPointerException because of calling `doneHandler.handle(null)`

mcweba (changes requested):

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueClient.java
ZhengXinCN: thanks fixed

mcweba (approved):

Kusig (approved):

PR #770, opened 2026-06-24
MSOP-7727 added Redisques batch queue support

ZhengXinCN:
added supports of batch queue from Redisques

### DIFF
--- a/gateleen-queue/README_queue.md
+++ b/gateleen-queue/README_queue.md
@@ -286,7 +286,7 @@ gateleen-packing packs incoming requests before they are queued. This feature ha
 
 A batched queue message must contain:
 
-Note: the non-payload proprieties are took from first queue item.
+Note: non-payload properties are taken from the first queue item. Headers are therefore expected to be homogeneous across the batch. `Content-Length` from the base item is removed because the merged payload size differs.
 
 ```json
 {
@@ -309,13 +309,14 @@ Note: the non-payload proprieties are took from first queue item.
 }
 ```
 When batchQueue is set to true, the queue processor treats payload as a JSON array of queue items.
+Batch queue processing supports JSON payloads only (each decoded item payload must be a JSON object).
 
 The processor:
 
-Uses the first item as the base request.
-Decodes each item’s Base64 payload.
-Adds all decoded payloads into one JSON array.
-Sends one HTTP request with the merged payload.
+* Uses the first item as the base request.
+* Decodes each item’s Base64 payload.
+* Adds all decoded payloads into one JSON array.
+* Sends one HTTP request with the merged payload.
 
 Example of source data:
 ```json
--- a/gateleen-queue/README_queue.md
+++ b/gateleen-queue/README_queue.md
@@ -317,13 +317,51 @@ Decodes each item’s Base64 payload.
 Adds all decoded payloads into one JSON array.
 Sends one HTTP request with the merged payload.
 
-Example merged HTTP payload:
+Example of source data:
 ```json
-[
-    { "key": "value" },
-    { "key": "value" }
+[ 
+  {
+    "method":"PUT",
+    "uri":"/playground/server/tests/exp/item_2",
+    "headers":[],
+    "payload":"eyJrZXkiOiAidmFsdWUifQ=="
+  },
+  {
+    "method":"PUT",
+    "uri":"/playground/server/tests/exp/item_2",
+    "headers":[],
+    "payload":"eyJrZXkiOiAidmFsdWUifQ=="
+  },
+  {
+    "method":"PUT",
+    "uri":"/playground/server/tests/exp/item_2",
+    "headers":[],
+    "payload":"eyJrZXkiOiAidmFsdWUifQ=="
+  }
 ]
 ```
+
+Example merged HTTP request:
+```json
+{
+  "method" : "PUT",
+  "uri" : "/playground/server/tests/exp/item_2",
+  "headers" : [ ],
+  "payload" : "W3sia2V5IjoidmFsdWUifSx7ImtleSI6InZhbHVlIn0seyJrZXkiOiJ2YWx1ZSJ9XQ"
+}
+```
+
+Example of merged payload, data from base64 encoded data "W3sia2V5IjoidmFsdWUifSx7ImtleSI6InZhbHVlIn0seyJrZXkiOiJ2YWx1ZSJ9XQ"
+
+```json
+[
+  {"key":"value"},
+  {"key":"value"},
+  {"key":"value"}
+]
+```
+
+
 
 ### API
 To check the current circuit states or close some or all circuits, the API handled by [QueueCircuitBreakerHttpRequestHandler](src/main/java/org/swisspush/gateleen/queue/queuing/circuitbreaker/api/QueueCircuitBreakerHttpRequestHandler.java) can be used.
--- a/gateleen-queue/README_queue.md
+++ b/gateleen-queue/README_queue.md
@@ -274,6 +274,57 @@ The configuration includes the activation / deactivation of the queue unlocking 
 
 **Note:** During the execution only 1 queue will be unlocked. Taken the configuration above, a queue is unlocked every 10s
 
+#### Redisques batch queue support
+Gateleen queue processing now supports batched queue messages produced by Redisques.
+
+Normally, Redisques sends queued items one by one to the queue processor. With batch queue support, Redisques can send multiple queued items in one event-bus message. Gateleen then merges these items and forwards them as one HTTP request.
+
+This is similar in purpose to packing, but it works at dequeue time.
+
+gateleen-packing packs incoming requests before they are queued. This feature handles queue items already dequeued by Redisques and lets Redisques deliver many queued items to Gateleen in one batch instead of one by one.
+##### Message format
+
+A batched queue message must contain:
+
+Note: the non-payload proprieties are took from first queue item.
+
+```json
+{
+  "batchQueue": true,
+  "queue": "my_queue",
+  "payload": [
+    {
+      "method": "PUT",
+      "uri": "/playground/server/tests/exp/item_2",
+      "headers": [],
+      "payload": "eyJrZXkiOiAidmFsdWUifQ=="
+    },
+    {
+      "method": "PUT",
+      "uri": "/playground/server/tests/exp/item_2",
+      "headers": [],
+      "payload": "eyJrZXkiOiAidmFsdWUifQ=="
+    }
+  ]
+}
+```
+When batchQueue is set to true, the queue processor treats payload as a JSON array of queue items.
+
+The processor:
+
+Uses the first item as the base request.
+Decodes each item’s Base64 payload.
+Adds all decoded payloads into one JSON array.
+Sends one HTTP request with the merged payload.
+
+Example merged HTTP payload:
+```json
+[
+    { "key": "value" },
+    { "key": "value" }
+]
+```
+
 ### API
 To check the current circuit states or close some or all circuits, the API handled by [QueueCircuitBreakerHttpRequestHandler](src/main/java/org/swisspush/gateleen/queue/queuing/circuitbreaker/api/QueueCircuitBreakerHttpRequestHandler.java) can be used.
 
--- a/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
+++ b/gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
@@ -3,6 +3,7 @@ package org.swisspush.gateleen.queue.queuing;
 import com.fasterxml.jackson.core.StreamReadConstraints;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
 import io.vertx.core.Vertx;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.eventbus.Message;
@@ -11,6 +12,7 @@ import io.vertx.core.http.HttpClient;
 import io.vertx.core.http.HttpClientRequest;
 import io.vertx.core.http.HttpClientResponse;
 import io.vertx.core.http.HttpMethod;
+import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.jackson.DatabindCodec;
 import org.slf4j.Logger;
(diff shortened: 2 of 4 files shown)


gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
```diff
@@ -86,19 +90,74 @@ public QueueProcessor(
...
-                HttpRequest queuedRequestTry = null;
-                JsonObject jsonRequest = parseStringToJsonObject(message.body().getString("payload"));
+            this.consumer = vertx.eventBus().consumer(getQueueProcessorAddress(), message -> {
+                HttpRequest queuedRequestTry;
+                JsonObject jsonRequest;
                 try {
-                    queuedRequestTry = new HttpRequest(jsonRequest);
+                    JsonObject messageBody = message.body();
+                    if (messageBody.getBoolean("batchQueue", false)) {
+                        log.debug("Batch queue message received");
```
Kusig: Such messages are useless if there is no indication what it is really.

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
```diff
@@ -86,19 +90,74 @@ public QueueProcessor(
...
+            JsonObject payloadObject = new JsonObject(new String(Base64Unit.decodeBase64Safe(reponseJsonObject.getString("payload"))));
+            payloadArray.add(payloadObject);
+        }
+        baseJsonObject.put(PAYLOAD, payloadArray.encode().getBytes());
+        return baseJsonObject;
+    }
+
     public void startQueueProcessing() {
         if (this.consumer == null || !this.consumer.isRegistered()) {
             log.info("about to register consumer to start queue processing");
```
Kusig: What consumer?

Kusig (changes requested):

Kusig:
- [ ]  Please await review from @mcweba

Kusig (approved):

mcweba:
Could you please add some documentation what this feature is about? How is the batchQueue property used, etc. It is quite hard to make a review when not knowing what's it about.

Have you took a look at the gateleen-packing module? Isn't this what you are trying to achieve?

ZhengXinCN:
@mcweba This looks like packing module, but it use for dequeue, vertx-redisques normally send queue items one by one, but now it supports send many queue items in a batch. yes I will add more documents

mcweba:
AI helped me to list potential problems with this new feature:

# Potential Problems with Batched Queue Feature
## 1. Silent Data Loss
   // Only first item's routing info used - rest silently discarded
   `JsonObject baseJsonObject = queueItems.getJsonObject(0).copy();`
- No warning if items have different URIs, methods, or headers
- No validation that items are compatible
- Debugging nightmare: "Why did my DELETE go to /users/1 as a PUT?"
## 2. Atomicity Issues
   Scenarios
   - Merged request fails -> **All N items** go back to queue and retry together
   - Backend partially processes -> Gateleen sees success, but some items silently failed
   - One bad payload -> Entire batch rejected

No per-item success/failure tracking!

## 3. Error Attribution Impossible
If the backend returns **400 Bad Request**:
- Which of the N payloads was invalid?
- Logs show one error, not which original queue item caused it
- Retry will fail repeatedly with same bad item in the batch

## 4. Payload Format Assumptions
`JsonObject payloadObject = new JsonObject(new String(Base64Unit.decodeBase64Safe(...)));`
`payloadArray.add(payloadObject);`
- **Assumes all payloads are JSON objects** - breaks if payload is a string, number, or array
- **Assumes payloads are valid JSON** - no try-catch around decode
- Binary payloads not supported
 
## 5. Expiry Check Flawed
`if (ExpiryCheckHandler.isExpired(queuedRequest.getHeaders(), jsonRequest.getLong(QueueClient.QUEUE_TIMESTAMP))) {`
- Uses **first item's timestamp** for expiry check
- If first item is fresh but others expired → expired items still get processed
- If first item expired → fresh items get discarded too

## 6. Timeout Inheritance
`String xTimeout = queuedRequest.getHeaders().get("x-timeout");`
- Uses **first item's x-timeout** header
- Batch of 100 items might need more time than single item's timeout
- No timeout scaling based on batch size

## 7. Circuit Breaker Distortion
`performCircuitBreakerActions(queueName, queuedRequest, SUCCESS, state);`
- One success/failure recorded for N items
- Statistics become meaningless (1 success could mean 1 or 1000 items)
- Circuit may stay closed despite high individual failure rate

## 8. Monitoring Inaccuracy
```java
if(monitoringHandler != null) {
    monitoringHandler.updateDequeue();  // Called once, not N times
}
```
- Dequeue count underreported by factor of batch size
- Throughput metrics completely wrong
- Capacity planning based on bad data

## 9. Size/Limit Risks
- No check on merged payload size
- Could exceed HTTP body limits
- Could cause OOM if batching thousands of items
- Backend might reject oversized requests

# 10. Tight Backend Coupling
- Backend **must** accept JSON arrays
- Backend must handle batch semantics (all-or-nothing vs partial)
- Changes how backend reports errors
- Not a drop-in optimization - requires backend changes
# 11. Queue Semantics Violation
Normal queue contract: "Each item processed independently"
Batched queue: Items are **coupled** - success/failure/retry happens together. This breaks assumptions code might have about queue behavior.

# 12. No Batch-Level Error Handling
```java
} catch (Exception exception) {
    log.error("Could not build batched request: {} ...", exception.getMessage());
    message.reply(new JsonObject().put(STATUS, ERROR).put(MESSAGE, exception.getMessage()));
    return;
}
```
- If one item's payload can't be decoded, **entire batch fails**
- No option to skip bad items and process the rest
  
# Summary
The feature trades correctness and observability for throughput. It's only safe when:
- All items are truly homogeneous (same destination, method, headers)
- Backend is designed for bulk operations
- You accept all-or-nothing batch semantics
- Monitoring/alerting accounts for batch sizes

Honestly, I see more problems than benefits with this solution. Consuming single queue items over the EventBus should be fast. Why don't you just implement a custom queue processor (maybe in Houston/Eagle) to consume multiple queue items and process them as a batch?

mcweba (changes requested):
See last comment

Kusig:
Thanks for helpful comments. For sure the technical issues must be taken in account.  We discussed in deep about different solutions for this dedicated use case.  For some cases it might really make sense where as for others it doesn't. And all the mentioned drawbacks must be properly described in the usage guide where no common solutions are possible. 

In the case of about 1000's of messages per second and multiple Pods all receiving them, the receiption by item makes simply no more sense and causes a lot of non necessary network traffic (yes, we hit limitations there). It simply as well get's to a maximum number of messages which is much lower that when sending in batches in dedicated use case. 


Just some first thoughts about:

10/11 and others) The client decides to consume this way and therefore must be changed and prepared anyway. If the client does not request it, it must not change at all the way the system behaves now. 

1/2/ and others) Logging of invalid item belongs to the consuming client. It is totally foreseen that all items in the batch are retried. If a client can't deal with that, it should simply not use it :-)

1) For sure the problem with different methods or other differences must be resolved and clarified.

5) If all messages in a batch are expired, the batch should expire at all and be discarded. Therefore, the last message is relevant, not the first one. 

6) The client can set the timeout on its registered listeners and therefore could deal perfectly with that itself and tune it as needed I think. 

About the proposed alternative, this simply makes no difference, most of the mentioned problems remain and the message bus load remains high which we wan't to get down as well with this for certain dedicated well matching use cases.

Proper documentation of the remaining drawbacks is mandatory of course.

ZhengXinCN:
@mcweba 

1. In the use case I have for this case, there only PUT. (Do I need group them by method?)

2. 
2.1 that is by design
2.2 that is depends on endpoint controller, just return non 20X will reject all
2.3 I didn't see any problem

3. this can be improved at backend side

4. for single (normal) also do the same, I didn't see different

5. I can add this

6. I think this need adjust when enqueue, not dequeue

7. Can be improved

8. Can be improved

9. there have a max limit at Redisques side, user need take care about this (documentation missing)

10. This is already by design, backend needs change

11. The related service use batch queues, need adjuest

12. Can be improved at backend

mcweba:
@ZhengXinCN 
1. Your specific case may only have PUT requests. When we add such features to Gateleen it should work for all cases. Generally spoken when working on gateleen or vertx-redisques we/you really have to take a look at the whole picture. Changes may work for a specific case but break everything else. When this happens in production, we are screwed. For this point I would suggest to only allow queue items with the same http method in a batch.

6. Timeout handling is not done on enqueue but on dequeue. When there is a queue item for long time in the queue the processing of the queue item has to be cancelled when it's expired.

As @Kusig mentioned, we need a really good documentation including drawbacks and potential problems.

Kusig:
About 1) There should be for sure only messages of the same HTTP method within a Badge. That said, if there is another one encountered (non PUT), it should proceed just normal as a single message as usual. This way as well clients could implement it per method properly and deal with it.  Don't know if this is easily feasible but from an architectural point of view it might be helpful. And of course properly documented.   As well the tests should assure that this works seamlessly.

ZhengXinCN:
@mcweba sorry, I didn't change anything yet, I mistakenly clicked the request review button

srudin (approved):

mcweba (changes requested):
I just put again to "Request changes" that it will not be merged after the latest approval, because the code has not changed yet

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueueProcessor.java
```diff
@@ -86,19 +90,74 @@ public QueueProcessor(
...
+                        }
+                    }
                 } catch (Exception exception) {
-                    log.error("Could not build request: {} error is {}", message.body().toString(), exception.getMessage());
+                    log.error("Could not build queue request: {} error is {}", message.body().toString(), exception.getMessage());
                     message.reply(new JsonObject().put(STATUS, ERROR).put(MESSAGE, exception.getMessage()));
                     return;
                 }
+
+                final JsonObject jsonRequestFinal = jsonRequest;
```
Kusig, line 198: Why this reassignment?  Worth a comment?

gateleen-queue/README_queue.md
```diff
@@ -274,6 +274,57 @@ The configuration includes the activation / deactivation of the queue unlocking
...
+When batchQueue is set to true, the queue processor treats payload as a JSON array of queue items.
+
+The processor:
+
+Uses the first item as the base request.
+Decodes each item’s Base64 payload.
+Adds all decoded payloads into one JSON array.
+Sends one HTTP request with the merged payload.
+
+Example merged HTTP payload:
```
Kusig: Is this example inline with the example above, I don't understand yet the relation and the comment just before which says that the decoded payload is put into one JSON array? Could you clarify or extend the docu here?

Kusig (changes requested):

ZhengXinCN:
@mcweba the different URI and METHOD problem is fixed in https://github.com/swisspost/vertx-redisques/pull/396

Kusig:
Claude Opus 4.8 Review results for the changed parts. Proably worth a look at those points:

Code Review — MSOP-7727-handle-the-batch-queue-items → develop

5 commits adding Redisques batch queue support: the QueueProcessor can now receive one event-bus message containing many queued items (batchQueue: true), filter expired ones, merge their payloads into a single JSON-array body, and forward one HTTP request.
Overall the feature is coherent, well-documented, and reuses existing helpers (ExpiryCheckHandler, JsonMultiMap, Base64Unit) correctly. A few real issues worth addressing before merge:

 🔴 Bugs / correctness
1. Platform-default charset (getBytes() / new String(...)) — QueueProcessor.mergeQueueResponses
JsonObject payloadObject = new JsonObject(new String(Base64Unit.decodeBase64Safe(...)));   // default charset
...
baseJsonObject.put(PAYLOAD, payloadArray.encode().getBytes());                              // default charset
Both rely on the JVM default charset, so any non-ASCII payload can be corrupted and behaviour becomes platform-dependent. Use StandardCharsets.UTF_8 explicitly in both places.

2. Merged request keeps the first item's headers — the base object is item[0], so Content-Length (if present) will describe item[0]'s body while the actual sent body is the larger merged array. Same for per-item headers like x-timeout, x-expire-after, retry headers — only the first item's are honored. At minimum strip Content-Length from the merged request; ideally document that batching assumes homogeneous headers.

3. Payload assumed to be a JSON object — new JsonObject(new String(decoded)) throws if any item's payload is a JSON array, scalar, empty, or non-JSON. The single-item path treats payload as opaque bytes, so batching silently imposes a new constraint. Because the exception makes the whole batch reply ERROR, one bad item causes the entire batch to be retried (all-or-nothing). Consider guarding/validating and documenting this limitation.
 


🟡 Minor / style
• Redundant expiry check: batch items are filtered in removeExpiredQueueItems, then executeQueuedRequest (line 339) re-checks expiry on the base item's timestamp. Harmless, but worth a comment.
• Double parsing: each item is parsed once in removeExpiredQueueItems and again in mergeQueueResponses. Minor inefficiency.
• Outer try/catch is now largely dead — both branches have their own try/catch that reply+return, so the outer catch rarely fires. Control flow could be simplified.
• Naming/typos: reponseJsonObject; new JsonObject(...)=new (missing space); method is called mergeQueueResponses / javadoc says "response" though these are requests. removeExpiredQueueItems could use queueItems.getString(i) instead of .getList().get(i).toString().
• README typos: "non-payload proprieties are took from first" → "properties are taken from the first"; a few missing bullet-list markers.
 

🟢 Tests
• testSuccessfulBatchedRequestResponse only asserts status=ok and that exactly one request is sent. It does not verify the merged body contents or the expired-item filtering — the two most interesting behaviors. Recommend adding a test that (a) captures the sent buffer and asserts the merged JSON array, and (b) includes an expired item and asserts it's dropped (and an all-expired batch replies OK with no request).
• RouteListingTest: adds haveSomeDelay() (100 ms Thread.sleep) after route (un)registration to stabilize flaky tests. Pragmatic but timing-based — a source of future flakiness/slowdown; fine as a stopgap.



Verdict
No blocking design problems, but I'd fix #1 (charset) and #2 (Content-Length/headers) before merge, and add real assertions to the batch test. Want me to apply the charset/header fixes and strengthen the test?

Kusig (changes requested):
Found myself nothing but I added the Claude Review results which are worth a check and some improvements I think.

Kusig (approved):

PR #778, opened 2026-07-14
SDCISA-23097 check and drop expired queue before enqueue it

ZhengXinCN:
Requests carrying an `X-Client-Timestamp` header (the timestamp set by the client when it originally issued the request) are checked for expiry before being enqueued. The expiration duration is taken from the `x-queue-expire-after` header, falling back to the `X-Expire-After` header when `x-queue-expire-after` is not present.

### DIFF
--- a/gateleen-packing/src/main/java/org/swisspush/gateleen/packing/PackingHandler.java
+++ b/gateleen-packing/src/main/java/org/swisspush/gateleen/packing/PackingHandler.java
@@ -45,11 +45,14 @@ public class PackingHandler {
     private MeterRegistry meterRegistry;
     private Counter packingRequestsSuccessCounter;
     private Counter packingRequestsFailCounter;
+    private Counter packingRequestsDropCounter;
 
     public static final String PACKING_REQUESTS_SUCCESS_COUNTER = "gateleen.packing.requests.success";
     public static final String PACKING_REQUESTS_SUCCESS_COUNTER_DESCRIPTION = "Amount of successfully packed requests processed";
     public static final String PACKING_REQUESTS_FAIL_COUNTER = "gateleen.packing.requests.fail";
     public static final String PACKING_REQUESTS_FAIL_COUNTER_DESCRIPTION = "Amount of failed packed requests processed";
+    public static final String PACKING_REQUESTS_DROP_COUNTER = "gateleen.packing.requests.drop";
+    public static final String PACKING_REQUESTS_DROP_COUNTER_DESCRIPTION = "Amount of dropped (expired) packed requests processed";
 
     /**
      * Constructs a new PackingHandler.
@@ -73,7 +76,7 @@ public class PackingHandler {
     }
 
     public PackingHandler(Vertx vertx, String queuePrefix, String redisquesAddress, String groupRequestHeader, PackingValidator validator, GateleenExceptionFactory exceptionFactory) {
-        this(vertx, queuePrefix, redisquesAddress, groupRequestHeader, validator, exceptionFactory, false);
+        this(vertx, queuePrefix, redisquesAddress, groupRequestHeader, validator, exceptionFactory, true);
     }
 
     public boolean isPacked(HttpServerRequest request) {
@@ -93,9 +96,13 @@ public class PackingHandler {
             packingRequestsFailCounter = Counter.builder(PACKING_REQUESTS_FAIL_COUNTER)
                     .description(PACKING_REQUESTS_FAIL_COUNTER_DESCRIPTION)
                     .register(meterRegistry);
+            packingRequestsDropCounter = Counter.builder(PACKING_REQUESTS_DROP_COUNTER)
+                    .description(PACKING_REQUESTS_DROP_COUNTER_DESCRIPTION)
+                    .register(meterRegistry);
         } else {
             packingRequestsSuccessCounter = null;
             packingRequestsFailCounter = null;
+            packingRequestsDropCounter = null;
         }
     }
 
@@ -139,6 +146,7 @@ public class PackingHandler {
 
                 if (!enqueueExpiredRequest && QueuingHandler.isRequestExpired(req.getHeaders())) {
                     requestLog.info("Dropping expired packed request '{}'", req.getUri());
+                    incrementDropCounter();
                     continue;
                 }
 
@@ -193,6 +201,12 @@ public class PackingHandler {
     private void incrementFailCounter() {
         if(packingRequestsFailCounter != null) {
             packingRequestsFailCounter.increment();
+        }
+    }
+
+    private void incrementDropCounter() {
+        if(packingRequestsDropCounter != null) {
+            packingRequestsDropCounter.increment();
         }
     }
 }
--- a/gateleen-packing/src/main/java/org/swisspush/gateleen/packing/PackingHandler.java
+++ b/gateleen-packing/src/main/java/org/swisspush/gateleen/packing/PackingHandler.java
@@ -41,6 +41,7 @@ public class PackingHandler {
     private final String queuePrefix;
     private final PackingValidator validator;
     private final GateleenExceptionFactory exceptionFactory;
+    private final boolean enqueueExpiredRequest;
     private MeterRegistry meterRegistry;
     private Counter packingRequestsSuccessCounter;
     private Counter packingRequestsFailCounter;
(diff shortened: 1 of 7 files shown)


gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueuingHandler.java
```diff
@@ -72,18 +75,39 @@ public QueuingHandler(
             HttpServerRequest request,
             RequestQueue requestQueue,
             QueueSplitter queueSplitter
+    ) {
+        this(vertx, redisProvider, request, requestQueue, queueSplitter, false);
```
mcweba: This breaks backward compatibility because it changes current behaviour. Before also expired queues will be enqueued. Default value must be `true`

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueuingHandler.java
```diff
@@ -72,18 +75,39 @@ public QueuingHandler(
...
+        this.enqueueExpiredRequest = enqueueExpiredRequest;
     }
 
     @Override
     public void handle(final Buffer buffer) {
         final String queue = request.headers().get(QUEUE_HEADER);
         final MultiMap headers = request.headers();
+
+        if (!enqueueExpiredRequest && isRequestExpired(headers)) {
+            // just skip this request, because the queue already expired
```
mcweba, line 133: Responding with a `202 Accepted` when the enqueuing is actually skipped is a bit misleading. In the logs you cannot see the difference. I would suggest do use another status code. Maybe `204 No Content`. With this at least we would see that the request was not added to the queue

gateleen-packing/src/main/java/org/swisspush/gateleen/packing/PackingHandler.java
```diff
@@ -130,6 +137,11 @@ public boolean handle(final HttpServerRequest request) {
                     req.getHeaders().remove(QueuingHandler.QUEUE_HEADER);
                 }
 
+                if (!enqueueExpiredRequest && QueuingHandler.isRequestExpired(req.getHeaders())) {
```
mcweba, line 164: Add metrics to count dropped packing requests.

```java
private Counter packingRequestsDropCounter;
```
Also update README_packing.md with documentation of the new metric and update unit tests

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueuingHandler.java
mcweba, line 133: Also add some logging like in PackingHandler like:
```
log.info("Dropping expired queued request '{}'", request.uri());
```

gateleen-queue/src/test/java/org/swisspush/gateleen/queue/queuing/QueuingHandlerTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for the {@link QueuingHandler#isRequestExpired(MultiMap)} method.
+ */
+public class QueuingHandlerTest {
+
+    private static final String CLIENT_TIMESTAMP_HEADER = "X-Client-Timestamp";
```
mcweba: Reuse existing constants:
```java
private static final String CLIENT_TIMESTAMP_HEADER = QueuingHandler.CLIENT_TIMESTAMP_HEADER;
private static final String EXPIRE_AFTER_HEADER = ExpiryCheckHandler.EXPIRE_AFTER_HEADER;
private static final String QUEUE_EXPIRE_AFTER_HEADER = ExpiryCheckHandler.QUEUE_EXPIRE_AFTER_HEADER;
```

mcweba (changes requested):

gateleen-queue/src/main/java/org/swisspush/gateleen/queue/queuing/QueuingHandler.java
ZhengXinCN: but will 204 blocks? if sender expected 202

mcweba: Do you have senders that actually expect 202? Our clients do not care whether they get a 200 or 202 or even a 204

Kusig: As we have as well providers which are not from ourself but from external companies, we could not say for sure if they accept 2xx  Codes in general here.  The question is mainy as well how it was documented.
However, changing the http status code would lead to some unforeseeable risk unfortunately if this was not clearly documented before. 
I will ask the main company if they accept 2xx codes in general.

mcweba: @Kusig thank you for checking. Another solution could also be to make the response configurable for dropped queue items. Default value would be a `202 Accepted` but could be configured to be something else.

srudin (approved):

ZhengXinCN:
@mcweba @Kusig I added configurable http status for rejected queue

Kusig (approved):

Kusig (changes requested):
Just as usual some final AI review to be checked please:

PR #778 — SDCISA-23097 check and drop expired queue before enqueue it

Adds pre-enqueue expiry checking so outdated requests aren't pushed to Houston. A request carrying X-Client-Timestamp is checked against x-queue-expire-after (fallback X-Expire-After); if expired and the handler is configured with enqueueExpiredRequest=false, it's dropped. Applies to both direct queuing (QueuingHandler) and packed sub-requests (PackingHandler). Adds a gateleen_packing_requests_drop_total metric, a configurable drop response status (default 202 Accepted), good docs, and thorough tests including a brand-new QueuingHandlerTest.

The design is sound: default behavior is unchanged (enqueueExpiredRequest=true on the backward-compatible constructors), the expiry helper is null/parse-safe, and partial drops in a packed batch keep the non-expired ones. Nicely done.

Issues & suggestions
1. Packed-batch response semantics: mixed drop+enqueue always returns 200, even if enqueues fail (medium)
In PackingHandler.handle, the response is decided synchronously right after the loop:

if (droppedExpiredRequests > 0 && enqueuedRequests == 0) {
    respondWith(request, expiredRequestStatusCode);
} else {
    respondWith(request, StatusCode.OK);
}
But the actual enqueue operations are async (eventBus callbacks that call incrementFailCounter() on failure). So respondWith runs before those complete — the response is already 200/202 regardless of whether the enqueues later succeed or fail. This mirrors the pre-existing behavior (it always responded OK before), so it's not a regression, but the new enqueuedRequests==0 branch makes the timing more relevant: if all items are dropped you respond 202 correctly, but if some are enqueued and then all fail async, you've still said 200. Worth a comment acknowledging the response is "accepted for processing," not "enqueued successfully."

2. isRequestExpired uses headers.isEmpty() short-circuit redundantly (low)
if (headers == null || headers.isEmpty()) return false;
String clientTimestamp = headers.get(CLIENT_TIMESTAMP_HEADER);
if (clientTimestamp == null) return false;
The isEmpty() check is harmless but redundant — the subsequent get(...) == null already covers the empty case. Not worth changing unless you want the micro-optimization; just noting it.

3. QueuingHandler.log is a non-static instance logger (nit → low)
private Logger log = LoggerFactory.getLogger(QueuingHandler.class);
Elsewhere in the codebase loggers are typically private static final. Since QueuingHandler is instantiated per request, this allocates/looks up a logger per instance. Prefer:

private static final Logger log = LoggerFactory.getLogger(QueuingHandler.class);
4. Header name constant vs. existing conventions (verify)
CLIENT_TIMESTAMP_HEADER = "X-Client-Timestamp" is defined fresh here. Please confirm there isn't already a constant for this header elsewhere (e.g. in ExpiryCheckHandler or a headers util) to avoid duplication — ExpiryCheckHandler already owns EXPIRE_AFTER_HEADER/QUEUE_EXPIRE_AFTER_HEADER, so the client-timestamp constant arguably belongs there too for cohesion.

5. Constructor proliferation (low — design)
QueuingHandler now has several overloads (+boolean, +StatusCode, +both). It's readable via telescoping delegation, but it's getting to the point where a small config object or builder might serve better long-term. Not blocking.

6. expiredRequestStatusCode null-guard is duplicated (nit)
Both PackingHandler and QueuingHandler do expiredRequestStatusCode == null ? StatusCode.ACCEPTED : expiredRequestStatusCode. Fine, but the MonitoringHandler-based QueuingHandler overload passes StatusCode.ACCEPTED explicitly while others default via the guard — consistent outcome, just two mechanisms achieving the same default. Minor.

Tests — strong
Genuinely good coverage:

QueuingHandlerTest (new): null headers, missing timestamp, unparseable timestamp, no expire headers, expired, not-yet-expired, x-queue-expire-after overriding X-Expire-After, infinite (-1) expiry, plus handle() returning 202 default and configured status with verifyNoInteractions(requestQueue). Excellent edge coverage of the helper.
PackingHandlerTest: drop-expired (default 202), drop with configured status, enqueue-expired-by-default, and the important mixed case (only expired dropped, valid one enqueued, 200 OK, drop+success counters both asserted).
Suggested additions:

A packed request where all sub-requests are expired → assert the drop counter increments for each and the response is 202 (the current all-dropped test has a single item; a 2+ all-expired case would lock the aggregate response).
The Thread.sleep(500) in several tests is a smell — it works but is timing-fragile on CI. Where possible prefer awaiting the metric via Awaitility (already used elsewhere in the codebase) instead of a fixed sleep.
Nits
README docs are clear and thorough (both queue and packing) — nice to see the behavior, default status, and override all documented, with an example.
requestLog (packing) vs log (queuing) naming differs across the two handlers — pre-existing, harmless.
Mockito.verifyNoInteractions(requestQueue) in the new tests is a good touch to prove nothing was enqueued.
Verdict
Approve with minor changes. Solid, well-scoped, backward-compatible, and well-tested. I'd suggest: make the QueuingHandler logger static final (#3), confirm there's no existing X-Client-Timestamp constant to reuse (#4), and add a short comment clarifying the packed-response timing (#1). The Thread.sleep → Awaitility swap would harden the tests against CI flakiness.

Kusig (approved):

PR #779, opened 2026-07-21
MSOP-8084 High number of EventBus Metric messages

ZhengXinCN:
optimize non-time-sensitive metrics, also allow to send metrics in batch to minimize eventbus load

### DIFF
--- a/gateleen-monitoring/README_monitoring.md
+++ b/gateleen-monitoring/README_monitoring.md
@@ -2,7 +2,7 @@
 This module provides monitoring functionality based on the [mod-metrics](https://github.com/swisspush/mod-metrics) module.
 
 ## Batch metric message (MonitoringHandler)
-`MonitoringHandler` supports sending non-time-sensitive metrics as one batch message on the event bus.
+`MonitoringHandler` supports sending gauge-like non-time-sensitive metrics as one batch message on the event bus.
 
 Enable this by using the constructor with `mergeMetricsIntoSingleMessage=true`:
 
@@ -18,11 +18,13 @@ Each entry in `metrics` uses the regular metric fields (`name`, `action`, `n`), 
 {
   "action": "batch",
   "metrics": [
-    { "name": "gateleen.requests.pending.count", "action": "set", "n": 2 },
+    { "name": "gateleen.hooks.listener.count", "action": "set", "n": 2 },
     { "name": "gateleen.queues.last.size", "action": "set", "n": 14 }
   ]
 }
 ```
+
+`requests.pending.count` is not merged into `batch`. It is buffered and flushed periodically as counter deltas (`inc`/`dec` with `n`) to preserve correct cross-instance aggregation.
 
 ## Request per Rule Monitoring
 Monitor incoming requests and routing rules matching these requests.
--- a/gateleen-monitoring/README_monitoring.md
+++ b/gateleen-monitoring/README_monitoring.md
@@ -1,6 +1,29 @@
 # gateleen-monitoring
 This module provides monitoring functionality based on the [mod-metrics](https://github.com/swisspush/mod-metrics) module.
 
+## Batch metric message (MonitoringHandler)
+`MonitoringHandler` supports sending non-time-sensitive metrics as one batch message on the event bus.
+
+Enable this by using the constructor with `mergeMetricsIntoSingleMessage=true`:
+
+`MonitoringHandler(Vertx vertx, ResourceStorage storage, String prefix, String requestPerRulePath, boolean mergeMetricsIntoSingleMessage)`
+
+When enabled, metrics are sent with:
+- `action = "batch"` (`MonitoringHandler.BATCH`)
+- `metrics = [...]` (`MonitoringHandler.METRICS_BATCH_ITEMS`)
+
+Each entry in `metrics` uses the regular metric fields (`name`, `action`, `n`), for example:
+
+```json
+{
+  "action": "batch",
+  "metrics": [
+    { "name": "gateleen.requests.pending.count", "action": "set", "n": 2 },
+    { "name": "gateleen.queues.last.size", "action": "set", "n": 14 }
+  ]
+}
+```
+
 ## Request per Rule Monitoring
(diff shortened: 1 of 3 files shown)


Kusig (changes requested):
Some AI generated feedback (Claude Opus 4.8) about this PR.  Probably you could run it yourself beforehand. I think some feedbacks are helfpul:


Summary
PR #779 — MSOP-8084 High number of EventBus Metric messages (gateleen producer side)

The change reduces EventBus load in MonitoringHandler two ways:

Buffering non-time-sensitive metrics locally and flushing them periodically (every 5s, on the existing queue-size timer) instead of emitting a message on every update. Pending-request count becomes a buffered delta (inc/dec with n); gauge-like metrics (last-used-queue-size, listener-count, route-count) are sent only when changed.
Optionally merging those gauge-like metrics into a single action=batch message (opt-in via new constructor flag mergeMetricsIntoSingleMessage, default false).
The consumer-side batch handling was refactored into handleMetricMessage(...) so single and batched items share one code path. Good docs, good backward compatibility (default off), extensive tests.

Issues & concerns
1. Concurrency: pending-request delta vs. gauge counters use different memory models (medium)
pendingRequestDelta uses AtomicLong with getAndSet — correct. But the gauge flushes mix reads/writes across two atomics non-atomically:

long current = listenerCount.get();
if (current != Long.MIN_VALUE && current != lastSentListenerCount.getAndSet(current)) { ... }
updateListenerCount() may run on a different thread than the flush timer. Between listenerCount.get() and lastSentListenerCount.getAndSet(current), a new update can arrive, and you could record lastSent = a value you haven't actually sent yet (you sent current, but store current as lastSent while a newer value already sits in listenerCount). The next flush then compares the newer value against current — usually fine, but the read-then-set isn't atomic, so a value can be skipped under races. In practice metrics are best-effort so impact is low, but worth a comment acknowledging the non-atomicity, or capturing the flush on the Vert.x context to guarantee single-threaded access.

2. Vert.x threading assumption for the flush (medium — verify)
flushBufferedMetrics() is invoked from a setPeriodic timer (Vert.x event-loop thread), but updateListenerCount/updateRoutesCount/updatePendingRequestCount/updateLastUsedQueueSizeInformation are public and may be called from arbitrary threads (request handlers, etc.). The atomics make individual reads/writes safe, but see #1. Confirm whether all callers are already on the same context; if so, most of the atomics could even be plain fields and #1 goes away. If not, the atomics are necessary but insufficient for the compound check. Worth an explicit decision + comment.

3. pendingLastUsedQueueName only keeps the last queue (behavior change — confirm intended)
Previously every updateLastUsedQueueSizeInformation(queue) call triggered an immediate redisques roundtrip + metric for that queue. Now only the most recently set queue survives to the next flush (getAndSet(null)), so if two different queues are updated within one 5s window, the first is dropped. For a single "last used queue size" gauge this is probably intended (you only care about the latest), but it is a semantic change — if downstream ever relied on per-queue updates, it's lost. Please confirm this is acceptable and ideally note it in the Javadoc.

4. Up-to-5s latency now added to all these metrics (low — confirm acceptable)
Every affected metric is now delayed up to NON_TIME_SENSITIVE_METRICS_PUBLISH_TIME (5s). The rename to "non-time-sensitive" signals this is intended, but pending-request count in particular used to be near-real-time. Confirm 5s is acceptable for alerting/dashboards built on requests.pending.count.

5. Field initialization moved into instance fields — check the second constructor path (low)
metricCache/lastDumps/metricLogger moved from local vars inside the constructor to instance fields. Good (they're now shared/consistent), but note the consumer lambda registers inside the 5-arg constructor only. Since all other constructors delegate to it, this is fine — just flagging that the EventBus consumer registration and the timer both now live on the delegated path. Verified OK from the diff.

6. handleMetricMessage batch items: no null/type guard (low)
for (int i = 0; i < items.size(); i++) {
    handleMetricMessage(items.getJsonObject(i));
}
getJsonObject(i) returns null (or throws ClassCastException) if an item isn't a JSON object; handleMetricMessage then NPEs on body.getString(...). Since the producer here is the only sender, risk is low, but the consumer is now also reachable by mod-metrics batches — a defensive null check/skip would harden it.

7. updateQueueCountInformation() + flushBufferedMetrics() both on the timer (low)
The periodic timer now does two things. updateQueueCountInformation() may itself emit metrics (unchanged), while flushBufferedMetrics emits the buffered ones. Fine, but the ordering means queue-count and buffered metrics can interleave. No correctness issue spotted; just noting the timer now has broader responsibility (the constant rename reflects that well).

Tests — strong
Excellent coverage: pending-delta buffering (inc/dec, no-op when unchanged), each gauge buffered-and-only-on-change, default-off flag across constructors, merged-batch content (sync + async metric in one batch), and the async last-used-queue-size path. The new @After closing Vert.x to stop the periodic timer leaking across tests is a nice catch.

Suggested additions:

A test asserting the two-different-queues-in-one-window behavior from #3 (locks the intended semantics).
A batch item that is null / not a JsonObject on the consumer path (#6).
Concurrency isn't easily unit-tested, but a comment documenting the threading model would substitute.
Nits
The consumer lambda cast (Handler<Message<JsonObject>>) is used in some tests and not others — cosmetic.
import java.util.*; wildcard already present; the new AtomicLong/AtomicReference imports are explicit — inconsistent but harmless.
Good renaming of QUEUE_SIZE_REFRESH_TIME → NON_TIME_SENSITIVE_METRICS_PUBLISH_TIME; the constant now covers more than queue size, so the name fits.
Fixed test typo tesExternalReceiver → testExternalReceiver — nice.
Verdict
Approve with minor changes. The design is sound, opt-in, backward-compatible, and well-tested. Before merge I'd want: confirmation on the threading model (#1/#2) with a clarifying comment, and confirmation that dropping all-but-last queue within a window (#3) and the added ≤5s latency (#4) are intended (a doc note each). #6 is a cheap hardening given the consumer is now shared with mod-metrics batches.

srudin (approved):

Kusig (approved):

PR #783, opened 2026-08-04
Sdcisa-20256 typescript hooks

NoApiManagement:
Create typescript hooks with features equivalent to the hooksjs


gateleen-hook-js/.gitignore
```diff
@@ -1,2 +1,4 @@
 node/
 node_modules/
+# Ignore package-lock.json to avoid corporate artefactory issues.
```
hiddenalpha: Typo `artefactory` -> [artifactory](https://jfrog.com/artifactory/)

gateleen-hook-js/README.md
hiddenalpha, line 1: That file looks a bit mis-placed. I assume there is some "totally obvious" reason it is in the hook-**js** module and not in the hook-**ts** one?

gateleen-hook-ts/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+#### `onClose(listener: () => void): () => void`
+
+Registers a callback invoked when EventBus closes. Returns cleanup function.
+
+```typescript
+eventBusService.onClose(() => console.log('Disconnected'));
+```
+
+## Types
```
hiddenalpha: Do we really want to duplicate our type definitions defined in code into the README?

gateleen-hook-ts/README.md
```diff
@@ -0,0 +1,293 @@
...
+- Smart hook refresh (only refreshes hooks nearing expiry)
+
+## Configuration
+
+### Socket Path
+
+By default, the EventBus connects to `/server/event/v1/sock`. To override:
+
+```typescript
+globalThis.GATELEEN_SOCKET_PATH = '/custom/event/socket';
```
hiddenalpha: Are we sure we want to use [global](https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil)s? Shouldn't that be something to pass-in via ctor or setter instead?

gateleen-hook-ts/README.md
```diff
@@ -0,0 +1,293 @@
...
+### Connection Timeout
+
+If EventBus takes longer than timeout to open, promise rejects:
+
+```typescript
+try {
+  await eventBusService.waitUntilOpen(3000); // 3 second timeout
+} catch (err) {
+  console.error('EventBus connection timeout');
+}
```
hiddenalpha: Does it really make sense to list a nearly identical code snippet twice? Couldn't we just mention the timeout case in the previous example and be done with it?

gateleen-hook-ts/README.md
```diff
@@ -0,0 +1,293 @@
...
+
+### Lint
+
+```bash
+npm run lint
+```
+
+## Browser Support
+
+- Modern browsers with ES2022 support
```
hiddenalpha, line 310: How many percent of users will we exclude by mandating "modern browsers"? (Is this the new "works-only-on-IE-6" facism the world had to cope with a while ago?)

However: For gateleen, I'm fine if it works for at least all the major browsers, in their respective LTS version ;)

The bare minimum: I assume this got tested that this works in at least **all** browsers we use throughout PaISA.

gateleen-hook-ts/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "test:ui": "vitest --ui",
+    "test:coverage": "vitest --coverage",
+    "lint": "tsc --noEmit",
+    "clean": "rm -rf dist node_modules",
+    "npm-link-for-local-use": "npm link",
+    "format:check": "prettier --check \"*.json\" \"src/**/*.{ts,html,css,scss,json}\"",
+    "format:fix": "prettier --write \"*.json\" \"src/**/*.{ts,html,css,scss,json}\""
+  },
+  "dependencies": {
+    "vertx3-eventbus-client": "^3.9.4"
```
hiddenalpha: `^3.9.4` are we sure we want to use versions which exchange themselves in between builds? (*see caret in there*). Same for dev-dependencies.

See also
- https://gitit.post.ch/projects/ISA/repos/notifications-api/pull-requests/7/overview?commentId=232144
- https://jira.post.ch/browse/SDCISA-352

gateleen-hook-ts/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+        </pluginManagement>
+    </build>
+
+    <profiles>
+        <profile>
+            <id>publicRepos</id>
+            <properties>
+                <nodeDownloadRoot>https://nodejs.org/dist/</nodeDownloadRoot>
+                <npmDownloadRoot>http://registry.npmjs.org/npm/-/</npmDownloadRoot>
+                <npmrcFile>.npmrc-public</npmrcFile>
```
hiddenalpha, line 146: Hmm? Isn't this already defined somewhere up in a parent pom? Do we really need to (re-)define it here again? Somehow I remember the term `publicRepos`.

(_correct me if I'm mistaken, and there is no such definition in any parent_)

gateleen-test/src/test/java/org/swisspush/gateleen/hookts/HookTsSteps.java
```diff
@@ -0,0 +1,134 @@
+/*
+ * ------------------------------------------------------------------------------------------------
+ * Copyright 2026 by Swiss Post, Information Technology Services
```
hiddenalpha: @mcweba @ljucam do we want such copyright notices in gateleen? Especially such which refer to Swiss Post? Is this a problem, as gatleeen itself seems to be apache 2.0 licensed.

gateleen-hook-ts/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+    "tabWidth": 2,
```
hiddenalpha, line 2: (*Likely this will just trigger never-ending discussions. So just take it as my opinion **and then ignore it***)

\<opinion>

tabWith "2"?!? (*its terrible using this for anything else than XML-like syntax*). As projects in our environment already refuse to use tabs (*which would allow everyone to set width to their own preference*), it should be **at least** "4" spaces IMHO. ([Some more reasoning](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation))

\</opinion>

gateleen-hook-ts/src/event-bus-service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      const result = await service.waitUntilOpen(100);
+      expect(result).toBeUndefined();
+    });
+
+    it('should reject on timeout', async () => {
+      vi.useFakeTimers();
+      try {
+        const service = new EventBusService();
+
+        const assertion = expect(service.waitUntilOpen(50)).rejects.toThrow(/EventBus open timeout/);
```
hiddenalpha: `toThrow(/EventBus open timeout/)` I assume this is a RegExp by intent?

gateleen-hook-ts/src/event-bus-service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  describe('onOpen', () => {
+    it('should register and call listener when EventBus opens', async () => {
+      const service = new EventBusService();
+      const listener = vi.fn();
+      const unsubscribe = service.onOpen(listener);
+
+      service.getEventBus().onopen?.();
+
+      await Promise.resolve();
```
hiddenalpha: `await Promise.resolve()`
WTF?!? - Should we at least explain (comment) why we do this?

gateleen-hook-ts/.gitignore
```diff
@@ -0,0 +1,16 @@
...
+*.log
+.idea/
+.vscode/
+*.swp
+*.swo
+*~
+.env
+.env.local
+target/
+# Ignore package-lock.json to avoid corparate artefactory issues.
```
hiddenalpha: Typo `artefactory` -> [artifactory](https://jfrog.com/artifactory/)

gateleen-hook-ts/src/hook-service.ts
```diff
@@ -0,0 +1,542 @@
...
+    try {
+      this.eventBus.unregisterHandler(managedRegistration.address, managedRegistration.registration.handler);
+    } catch (err) {
+      console.warn(`Failed to unbind local EventBus handler for hook ${managedRegistration.registration.id}:`, err);
+    } finally {
+      managedRegistration.handlerBound = false;
+    }
+  }
+
+  private async fetch<T>(url: string): Promise<T | null> {
```
hiddenalpha: `private async fetch(url)`: Are we sure we want to name a method exactly the same as a [global](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch)?

gateleen-hook-ts/src/hook-service.ts
```diff
@@ -0,0 +1,542 @@
...
+  registration: Registration<any>;
+  address: string;
+  expiresAt: number;
+  disposed: boolean;
+  handlerBound: boolean;
+  inFlightRegistration?: Promise<void>;
+}
+
+/**
+/**
```
hiddenalpha, line 509:
```
/**
/**
```
?

gateleen-hook-ts/README.md
NoApiManagement: "How many percent of users will we exclude by mandating "modern browsers"?" 0% 
if I am reading https://compat-table.github.io/compat-table/es2016plus/ correctly, all the major browsers supported that by end of 2022 I think.

gateleen-hook-ts/pom.xml
NoApiManagement: It's also like that in hooks-js gateleen-hook-js/pom.xml:209 
nodeDownloadRoot only exists in hooks-js and hooks-ts

gateleen-test/src/test/java/org/swisspush/gateleen/hookts/HookTsSteps.java
NoApiManagement: gateleen-hook-js has a few... 
will remove them in the hook-ts

gateleen-hook-ts/README.md
NoApiManagement: Will remove it. 
My first thought was that someone using this could only read the readme and not look at the code.

gateleen-hook-ts/README.md
NoApiManagement: will remove it

gateleen-test/src/test/java/org/swisspush/gateleen/hookts/HookTsSteps.java
mcweba: We have only very few of such copyright entries in Gateleen. I would also propose to remove them

gateleen-hook-ts/src/event-bus-service.spec.ts
NoApiManagement: I changed it, since the time is being controlled anyway I can have the whole string anyway.
What I just learned is that toThrow("EventBus open timeout"); would have done the same (seems like its contains or starts with not equals

gateleen-hook-ts/src/event-bus-service.spec.ts
NoApiManagement: Thanks! that comment made me find a few flaws in the testing... I added a comment in the code.

Flush the microtask queue (a lighter weight version of await new Promise(r => setTimeout(r, 0)) ).

gateleen-hook-ts/README.md
NoApiManagement: I removed the code snippet, left the section that it can time out.

gateleen-hook-js/README.md
NoApiManagement: 2 "Reasons", but I moved it into the hooks-ts as they where not so good anyway.
hooks-js is before hooks-ts, so if they failed the instructions would have been helpful there already.
hooks-ts already has a README.md, so if you don't run the tests manually but let an agent run them then it takes a tiny bit more tokens...

But as if in the future we ever remove the JS part then it makes sense to have it in the TS part anyway...

gateleen-hook-ts/.prettierrc
NoApiManagement: I see the argument for more than 2, I find it compelling.
2 is the default for .prettierrc , https://prettier.io/docs/options.html#tab-width .
I don't mind changing it, but in a separate PR.

hiddenalpha (approved):

hiddenalpha:
Merging on behalf of *NoApiManagement*, as he seems not to have a merge button.

PR #787, opened 2026-08-27
Fixed withFullUrl for listeners

ZhengXinCN:
Fixed the withFullUrl logic for listeners. Seems currently it is only working for routes.


gateleen-hook/README_hook.md
```diff
@@ -186,15 +187,28 @@ PUT http://myserver:7012/gateleen/everything/_hooks/listeners/http/myexample
...
     ],
     "headersFilter": "x-foo: (A|B)"
 }
 ```
 
+With `forcedTargetPath`, a matching request such as
+`/gateleen/from/services/orders/my/data/item-123` matches the listener filter.
+Without `forcedTargetPath`, the complete suffix
+`/orders/my/data` is forwarded, resulting in
+`/gateleen/to/services/orders/my/data`. With
```
srudin: Not sure the example is fully correct and meaningful:
- The filter is "/gateleen/from/services/[^/]+/my/data" so maybe we should make an example like "/gateleen/from/services/order-123/my/data" because "/gateleen/from/services/orders/my/data/item-123" will not match, right?
- The suffix in your example should be "/orders/my/data/item-123", no? So in my example it would become "/order-123/my/data". The 123 is important here because that's the reason we're doing this.

srudin (approved):

gateleen-hook/README_hook.md
ZhengXinCN: @srudin updated, can you check again?

mcweba (changes requested):
See comments made before

mcweba (changes requested):
See my last comment. I like to discuss this first before doing another code review for the forcedTargetPath feature

gateleen-hook/docs/fullUrl.svg
srudin, line 1: Maybe also add what happens with routes? Because it can be used for both, not only listeners, as the title suggests.

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
```diff
@@ -872,20 +872,17 @@ private void callListener(RoutingContext ctx, final Buffer buffer, final List<Li
...
-            // internal
             if (listener.getHook().getDestination().startsWith("/")) {
-                targetUri = listener.getListener() + path;
+                targetUri = listener.getHook().isFullUrl() ? listener.getHook().getDestination() : listener.getListener() + path;
                 log.debug(" > internal target: {}", targetUri);
-            }
-            // external
-            else {
+            } else {
                 targetUri = hookRootUri + LISTENER_HOOK_TARGET_PATH + listener.getListener() + path;
```
srudin, line 886: I think this should be:
```
if (listener.getHook().isFullUrl()) {
    targetUri = listener.getHook().getDestination();
} else if (listener.getHook().getDestination().startsWith("/")) {
... else ...
}
```
Because otherwise fullUrl won't work for absolute destinations. Not something we use now, but maybe one day...

srudin (changes requested):

gateleen-hook/src/main/java/org/swisspush/gateleen/hook/HookHandler.java
ZhengXinCN: impossible, a url must starts with a /, in case Fullurl, the Destination already starts with /

ZhengXinCN: impossible, a url must starts with a /, in case Fullurl, the Destination already starts with /

if give a Destination:  server/test waith Fullurl = ture
what should be the targetUri? becasue it is FullUrl, I can't have hookRootUri + LISTENER_HOOK_TARGET_PATH

srudin (approved):

mcweba (approved):

PR #788, opened 2026-08-27
SDCISA-23073 prevent chunked empty body

ZhengXinCN:
if there is an empty body response, the chunked flag will not set


srudin (approved):

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -662,13 +662,17 @@ private void onUpstreamResponse(HttpClientResponse rsp, RequestCtx ctx) {
...
-            ctx.dnRsp.setChunked(true);
-        }
-
-        final LoggingWriteStream loggingWriteStream = new LoggingWriteStream(ctx.dnRsp, ctx.loggingHandler, false);
+        // Do not forward transfer framing. Enable chunked encoding only if the upstream
+        // response actually yields payload bytes; otherwise Vert.x would emit an empty
+        // chunked response.
+        boolean upstreamResponseIsChunked = ctx.dnRsp.headers()
+                .contains(HttpHeaders.TRANSFER_ENCODING, "chunked", true);
+        ctx.dnRsp.headers().remove(HttpHeaders.TRANSFER_ENCODING);
```
hiddenalpha, line 670:
```
upstreamResponseIsChunked = ctx.dnRsp.headers()
//                        here -^^
```
Comment states we inspect **UP**stream for chunked encoding. But why the code actually inspects the **DOWN**stream response?

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderChunkedResponseTest.java
```diff
@@ -0,0 +1,264 @@
...
+        Async backendReady = ctx.async();
+        int[] port = new int[1];
+        backend = vertx.createHttpServer();
+        backend.requestHandler(req -> responder.handle(req.response()));
+        backend.listen(0, ctx.asyncAssertSuccess(server -> {
+            port[0] = server.actualPort();
+            backendReady.complete();
+        }));
+        backendReady.awaitSuccess(5000);
+        return port[0];
```
hiddenalpha: How can we return `port[0]`? Isn't it always zero, as our listen callback is **async** and only will set the contents later? (*when we've already returned the value*)

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/ConditionalChunkedResponseWriteStream.java
```diff
@@ -0,0 +1,89 @@
...
+import io.vertx.core.http.HttpServerResponse;
+import io.vertx.core.streams.WriteStream;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Enables chunked transfer encoding immediately before forwarding the first
+ * non-empty response buffer.
+ */
+final class ConditionalChunkedResponseWriteStream implements WriteStream<Buffer> {
```
mcweba: # Duplicates AutomaticChunkedTransfer in the same package, with divergent semantics

**AutomaticChunkedTransfer** (same package) already implements lazy setChunked for the request direction, with a documented rationale, publishError fallback chain, and a retained exceptionHandler reference. The new class solves the mirror problem with:

- a different empty-buffer rule (**AutomaticChunkedTransfer** sets chunked on **any** first buffer; the new one only on non-empty),
- no error propagation at all,
- no dbgHint for log correlation.

**Risk:** Two classes, one package, same concern, silently different behaviour. The next person to touch transfer framing has to diff them to know which rule applies in which direction. The divergence in finding 2 is a direct consequence.

**Smallest Fix:** Align the empty-buffer rule with AutomaticChunkedTransfer (see finding 2), and cross-reference it in the Javadoc:
```
/**
 * Response-side counterpart to {@link AutomaticChunkedTransfer}.
 */
```

ZhengXinCN: AutomaticChunkedTransfer  works for Request only, so now I renamed it to AutomaticChunkedRequestTransfer , what I added renamed to AutomaticChunkedResponseTransfer. and all shared code moved into a base class

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/AutomaticChunkedRequestTransfer.java
```diff
@@ -37,33 +33,33 @@
...
-        assert !isNullOrEmpty(dbgHint) : "An empty dbgHint is worth NOTHING!";
         this.vertx = vertx;
-        this.delegate = delegate;
-        this.dbgHint = dbgHint;
+    }
+
+    @Override
+    protected void setDelegateChunked() {
+        // avoid multiple calls due to a 'syncronized' block in HttpClient's implementation
+        delegate.setChunked(true);
```
hiddenalpha, line 50: > // avoid multiple calls due to a 'syncronized' block in HttpClient's implementation

What does this comment try to tell me? Looks like that comment should be somewhere else? Here it looks mis-placed somehow?

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/AutomaticChunkedRequestTransfer.java
```diff
@@ -37,33 +33,33 @@
...
+    @Override
+    protected boolean isDelegateChunked() {
+        return delegate.isChunked();
     }
 
     private void write_(Buffer data, Handler<AsyncResult<Void>> handler) {
         /* only now we know for sure that there IS a body. */
         Future.<Void>succeededFuture().<Void>compose((Void nil) -> {
             if (firstBuffer.getAndSet(false)) {
-                // avoid multiple calls due to a 'syncronized' block in HttpClient's implementation
```
hiddenalpha:
```
  if (firstBuffer.getAndSet(false)) {
      // avoid multiple calls due to a 'syncronized' block in HttpClient's implementation
```

Eh :D Look at that. I assume that comment I just mentioned before should be here, as it refers to `firstBuffer.getAndSet(false)`.

hiddenalpha, line 62: Ah, THATs likely the place where that previously mentioned comment should go. As it describes what `firstBuffer.getAndSet(false)` is for.

gateleen-routing/src/test/java/org/swisspush/gateleen/routing/ForwarderChunkedResponseTest.java
```diff
@@ -139,15 +140,19 @@ public void testContentLengthUpstreamResponse_downstreamIsNotChunked(TestContext
...
         backend = vertx.createHttpServer();
         backend.requestHandler(req -> responder.handle(req.response()));
         backend.listen(0, ctx.asyncAssertSuccess(server -> {
-            port[0] = server.actualPort();
+            port.set(server.actualPort());
             backendReady.complete();
         }));
         backendReady.awaitSuccess(5000);
-        return port[0];
+        return port.get();
```
hiddenalpha, line 155: This does not solve the mentioned issue at all. Still, we seem to `return port.get()` **BEFORE** we call `port.set(server.actualPort())`.

ZhengXinCN: sorry, I wasn't take care too much because it is tests   code

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -658,17 +658,21 @@ private void onUpstreamResponse(HttpClientResponse rsp, RequestCtx ctx) {
         // Add received headers to original request but remove headers that should not get forwarded.
         MultiMap headersToForward = ctx.upRes.headers();
         headersToForward = removeNonForwardHeaders(headersToForward);
+        // Do not forward transfer framing. Enable chunked encoding only if the upstream
+        // response actually yields payload bytes; otherwise Vert.x would emit an empty
+        // chunked response.
+        boolean upstreamResponseIsChunked = headersToForward.contains(HttpHeaders.TRANSFER_ENCODING, "chunked", true);
```
mcweba, line 664: ### 1. 🔴 `gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java:664-670` - Valid transfer-coding lists lose both framing and coding metadata

**Code:**
```java
boolean upstreamResponseIsChunked = headersToForward.contains(HttpHeaders.TRANSFER_ENCODING, "chunked", true);
HttpHeaderUtil.mergeHeaders(ctx.dnRsp.headers(), headersToForward, ctx.targetUri);
if (ctx.profileHeaderMap != null && !ctx.profileHeaderMap.isEmpty()) {
    HttpHeaderUtil.mergeHeaders(ctx.dnRsp.headers(), MultiMap.caseInsensitiveMultiMap().addAll(ctx.profileHeaderMap), ctx.targetUri);
}
ctx.dnRsp.headers().remove(HttpHeaders.TRANSFER_ENCODING);
```

**Problem:** `MultiMap.contains(name, value, true)` compares a complete header value. A legal `Transfer-Encoding: gzip, chunked` field is therefore not recognized as chunked. The code then removes the entire field. The response wrapper is configured with `false`, so it does not add downstream chunked framing. A non-empty body with no `Content-Length` reaches `HttpServerResponse.write()` without framing. Vert.x rejects exactly this HTTP/1.1 state before sending the body. A close-delimited non-chunked coding such as `Transfer-Encoding: gzip` is also discarded, causing a recipient to treat encoded bytes as an identity representation.

**Impact:** Requests through a backend that uses a legal transfer-coding chain fail while forwarding the first body buffer, or return bytes with the wrong representation semantics. The intended fix for empty chunked bodies turns valid upstream framing into an invalid downstream response.

**Challenge:** This would be harmless if the upstream client normalized every legal `Transfer-Encoding` list into a separate header value containing only `chunked`, or if the gateway contract prohibited all transfer codings other than a lone `chunked`. The forwarding path has neither normalization nor validation. It uses the raw `HttpClientResponse.headers()` map, and `removeNonForwardHeaders()` only removes connection-specific fields.

**Evidence:** Proven - `testChainedTransferEncodingIsRejectedInsteadOfDroppingFraming` uses a raw origin response with a valid gzip-compressed, chunked body and a real downstream Vert.x server/client. It fails on the current branch because the client receives `200` instead of `502`, then the gateway logs `IllegalStateException: You must set the Content-Length header to be the total size of the message body BEFORE sending any data if you are not using HTTP chunked encoding.` from `AutomaticChunkedResponseTransfer.write`. This follows the same path traced above: `Forwarder` derives `useChunkedTransfer` only from the exact-value lookup, removes the header, and passes `false` to `AutomaticChunkedResponseTransfer`, which does not call `setChunked(true)`.

```java
/**
 * A proxy must not silently discard transfer codings it cannot preserve. In particular,
 * {@code chunked} is legal as the final token of a transfer-coding list, rather than only
 * as the complete header value. The proxy should reject this unsupported representation
 * with 502 instead of attempting an unframed downstream write.
 */
@Test
public void testChainedTransferEncodingIsRejectedInsteadOfDroppingFraming(TestContext ctx) throws IOException {
    Async async = ctx.async();
    Buffer gzipBody = gzip("hello");
    io.vertx.core.net.NetServer backend = vertx.createNetServer();
    backend.connectHandler(socket -> socket.handler(request -> {
        Buffer response = Buffer.buffer("HTTP/1.1 200 OK\r\n" +
                "Transfer-Encoding: gzip, chunked\r\n" +
                "Connection: close\r\n" +
                "\r\n");
        response.appendString(Integer.toHexString(gzipBody.length()));
        response.appendString("\r\n");
        response.appendBuffer(gzipBody);
        response.appendString("\r\n0\r\n\r\n");
        socket.write(response);
    }));

    backend.listen(0).onSuccess(origin -> {
        Router router = Router.router(vertx);
        router.route().handler(forwarderFor(origin.actualPort()));
        vertx.createHttpServer().requestHandler(router).listen(0).onSuccess(gateway ->
                vertx.createHttpClient()
                        .request(HttpMethod.GET, gateway.actualPort(), "localhost", REQUEST_URI)
                        .compose(request -> request.send())
                        .compose(response -> {
                            ctx.assertEquals(502, response.statusCode(),
                                    "Unsupported transfer codings must produce a gateway error");
                            return response.body();
                        })
                        .onComplete(ctx.asyncAssertSuccess(body -> async.complete()))
        ).onFailure(ctx::fail);
    }).onFailure(ctx::fail);

    async.awaitSuccess(3000);
}

private static Buffer gzip(String body) throws IOException {
    ByteArrayOutputStream output = new ByteArrayOutputStream();
    try (GZIPOutputStream gzip = new GZIPOutputStream(output)) {
        gzip.write(body.getBytes());
    }
    return Buffer.buffer(output.toByteArray());
}
```

**Smallest Fix:**
```java
List<String> transferEncodings = headersToForward.getAll(HttpHeaders.TRANSFER_ENCODING);
boolean upstreamResponseIsChunked = transferEncodings.stream()
        .flatMap(value -> Arrays.stream(value.split(",")))
        .map(String::trim)
        .anyMatch(value -> "chunked".equalsIgnoreCase(value));
if (transferEncodings.stream().anyMatch(value -> !"chunked".equalsIgnoreCase(value.trim()))) {
    respondError(ctx.dnReq, BAD_GATEWAY);
    return;
}
ctx.dnRsp.headers().remove(HttpHeaders.TRANSFER_ENCODING);
```

This deliberately rejects transfer codings the proxy does not implement instead of silently removing their meaning. If support for chained codings is required, decode the non-chunked coding or preserve it while re-chunking; do not apply the rejection branch.

mcweba (changes requested):

mcweba (approved):

PR #795, opened 2026-09-21
Msop 9475 retry on bad gateway

Kusig:
### Summary
Adds an optional, globally configurable retry for forwarded requests that fail while waiting for the upstream response (the path that returns 502). Disabled by default; only retries requests that are safe to replay.

### What changed
• New system property for the total attempts (initial try + retries):   org.swisspush.gateleen.routing.forwarder.maxForwardAttempts (default 1 = no retry; invalid/<1 values clamp to 1).
• A retry runs only when all hold: budget left, idempotent method (GET/HEAD/PUT/DELETE/OPTIONS/TRACE; never POST/PATCH), buffered body (streamed bodies can't be replayed), and nothing written downstream yet.
• Extracted request initiation into a re-callable sendToTarget(ctx); added a per-request attempt counter. Retries are sequential and stop on the first upstream response.
 
### Testing
New ForwarderRetryTest (13 tests): property parsing/sanitizing, no-retry default, budget 3 always-failing → 3 attempts + 502, recovery on 2nd/3rd try → 200, and no-retry for POST and streamed bodies. Existing forwarder/router tests re-run green.

### Compatibility
Fully backward compatible — with the property unset, behavior is unchanged.
Example (one retry): -Dorg.swisspush.gateleen.routing.forwarder.maxForwardAttempts=2


gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -92,6 +93,65 @@ public class Forwarder extends AbstractForwarder {
...
+     * Values below {@code 1} or non-numeric values are ignored and fall back to {@code 1}.
+     * <p>
+     * The value is resolved once per {@link Forwarder} instance (in the constructor via
+     * {@link #resolveMaxForwardAttempts()}); it is "global" in the sense that every forwarder reads
+     * the very same JVM-wide system property.
+     * <p>
+     * Retries only happen for idempotent request methods (see {@link #isIdempotent(HttpMethod)})
+     * whose body is already buffered in memory, because a streamed request body cannot be replayed.
+     */
+    public static final String MAX_FORWARD_ATTEMPTS_PROPERTY = "org.swisspush.gateleen.routing.forwarder.maxForwardAttempts";
```
mcweba, line 118: ### Significant `gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java:112` - Global retry property is snapshotted per forwarder

**Action:** Required

**Evidence:** Proven - `ForwarderRetryTest.testRetryBudgetIsSnapshottedWhenForwarderIsConstructed` constructs a forwarder with a budget of `1`, changes the property to `3`, and observes one attempt from the existing forwarder followed by three attempts from a newly constructed one. `mvn -pl gateleen-routing -am -Dtest=ForwarderRetryTest -Dsurefire.failIfNoSpecifiedTests=false test` passed 14 tests. The code path confirms the mechanism: `maxForwardAttempts` is initialized once from `resolveMaxForwardAttempts()` at lines 112-153. Router forwarders are rebuilt on routing-rule changes (`Router.java:348-380`), but hook `Route` forwarders are retained when hook configuration is unchanged (`HookHandler.java:1805-1824`).

**Problem:** The advertised global property is read only during each `Forwarder` construction. A runtime property change therefore leaves existing hook routes on their old retry budget while subsequently created forwarders use the new value.

**Impact:** A single running Gateleen instance can apply inconsistent retry policies despite the property being documented as globally configurable.

**Smallest Fix:** Resolve the property when evaluating a retry, or explicitly define it as startup-only configuration and ensure the documentation and configuration lifecycle reflect that constraint.

**Evidence**
```java
    /**
     * Demonstrates that the retry budget is captured when a Forwarder is constructed rather than
     * read from the global property for each request. This is evidence for the configuration
     * lifecycle concern: changing the property leaves existing forwarders on their previous budget.
     */
    @Test
    public void testRetryBudgetIsSnapshottedWhenForwarderIsConstructed(TestContext ctx) {
        AtomicInteger hits = new AtomicInteger();
        int backendPort = startCountingBackend(ctx, hits, Integer.MAX_VALUE);

        System.setProperty(Forwarder.MAX_FORWARD_ATTEMPTS_PROPERTY, "1");
        Rule existingRule = buildRule(backendPort);
        Forwarder existingForwarder = buildForwarder(existingRule,
                vertx.createHttpClient(existingRule.buildHttpClientOptions()));

        System.setProperty(Forwarder.MAX_FORWARD_ATTEMPTS_PROPERTY, "3");
        int existingStatus = forwardAndAwaitStatus(ctx, existingForwarder);

        ctx.assertEquals(StatusCode.BAD_GATEWAY.getStatusCode(), existingStatus);
        ctx.assertEquals(1, hits.get(), "Existing forwarder retained the budget from construction");

        Rule newRule = buildRule(backendPort);
        Forwarder newForwarder = buildForwarder(newRule,
                vertx.createHttpClient(newRule.buildHttpClientOptions()));
        int newStatus = forwardAndAwaitStatus(ctx, newForwarder);

        ctx.assertEquals(StatusCode.BAD_GATEWAY.getStatusCode(), newStatus);
        ctx.assertEquals(4, hits.get(), "New forwarder used the updated budget of 3 attempts");
    }
```

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
mcweba, line 544: ### Significant `gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java:545-546, 694, 680-684` - Retries multiply the configured request timeout

**Action:** Required

**Evidence:** Traced - `handleRequest` captures `rule.getTimeout()` or `x-timeout` once as `ctx.timeoutMs` (`Forwarder.java:488-497`). Every call to `sendToTarget` creates a fresh pool-wait timer for the full value (`545-546`) and applies the full value as the upstream request's idle timeout (`694`). A failed response calls `sendToTarget(ctx)` again (`680-684`) without reducing or otherwise enforcing a request-wide deadline. The new retry tests assert attempt counts and outcomes but do not assert the elapsed time or timeout budget.

**Problem:** A configured timeout no longer bounds the total forwarded request lifetime when retries are enabled. With a retry budget of `N`, requests can wait up to roughly `N * timeout` while receiving no upstream response, plus connection and scheduling overhead.

**Impact:** Slow or unhealthy upstreams retain downstream connections and pool capacity substantially longer than the route or caller configured. For hooks, whose documented timeout is a request timeout, this can violate latency limits and amplify load during backend failure.

**Smallest Fix:** Store a request-wide deadline in `RequestCtx` when forwarding starts. Before each attempt, derive the remaining duration; use that value for both the pool-wait timer and `idleTimeout`, and skip retrying when no time remains. Add a focused test with a short timeout and a failing backend that verifies the final response occurs within one end-to-end budget rather than one budget per attempt.

Kusig: Good catch, I didn't had this important detail in my radar :-(

gateleen-routing/src/main/java/org/swisspush/gateleen/routing/Forwarder.java
```diff
@@ -496,6 +507,19 @@ private void handleRequest(final HttpServerRequest req, final Buffer bodyData, f
                 req, log, targetUri, startTime, timerSample, profileHeaderMap, loggingHandler,
                 afterHandler, timeout, timeoutMs, uniqueId, authHeader.orElse(null), bodyData);
 
+        /*
+         * Establish a single request-wide deadline that bounds the retry span of the forwarded
+         * request. The FIRST attempt still runs on the full configured timeout (see
```
mcweba: You already mentioned that the first attempt still runs on the full configured timeout.

See the (AI generated) proposal how this could be improved:

**Smallest Fix:** Use the remaining request-wide deadline for both the pool-wait timer and `idleTimeout` on every attempt, including the first. Preserve any intentional pre-existing per-phase semantics only if the route timeout is explicitly documented as such.

## Implementation Proposal

Apply `remainingTimeoutMs(ctx)` to the pool-wait timer for every attempt, rather than retaining the configured timeout for the first attempt:

```diff
private void sendToTarget(RequestCtx ctx) {
    final AtomicBoolean responded = new AtomicBoolean(false);

-   final long attemptTimeoutMs = perAttemptTimeoutMs(ctx);
+   // Pool waiting consumes the request-wide timeout on every attempt.
+   final long attemptTimeoutMs = remainingTimeoutMs(ctx);

    final long poolWaitTimerId = ctx.timeoutMs > 0
        ? vertx.setTimer(attemptTimeoutMs, id -> {
            // Existing timeout response handling.
        })
        : -1;
}
```

Recalculate the budget after a connection is acquired, because pool waiting may have consumed most of it before the upstream idle timeout is set:

```diff
-ctx.upReq.idleTimeout(perAttemptTimeoutMs(ctx));
+// Pool acquisition may have consumed most of the request-wide timeout.
+ctx.upReq.idleTimeout(remainingTimeoutMs(ctx));
```

`remainingTimeoutMs(ctx)` already preserves the existing no-timeout behavior (`timeoutMs <= 0`) and clamps an expired configured deadline to `1` ms, avoiding Vert.x's special `0` value, which disables the idle timeout. With these call-site changes, `perAttemptTimeoutMs` is no longer required and can be removed.

Add a regression test that occupies the route's sole pool connection until nearly the configured timeout expires, then releases it to a backend that never responds. Assert that the final downstream response occurs within approximately one configured timeout window, rather than a pool-wait window plus a fresh upstream-idle window.

